PR #2, opened 2021-09-21
Some clean up for VMs template

nocturnalastro:


roles/create_vms/tasks/main.yml
```diff
@@ -4,8 +4,8 @@
   set_fact:
     kvm_nodes: "{{ kvm_nodes | default([]) + [{
         'name': item,
+        'uuid': hostvars[item]['uuid'] | default(false),
```
mkochanowski, line 7: I think a more intuitive default value would be an empty string. We should not mix different types (uuid string, boolean) for the same variable (`kvm_nodes[*]['uuid']`).

mkochanowski (changes requested):

crwr45 (approved):

mkochanowski (approved):

PR #3, opened 2021-09-21
Missing day2 when in deploy_day2_workers

nocturnalastro:


mkochanowski (approved):

deploy_day2_workers.yml
```diff
@@ -26,4 +26,5 @@
 - hosts: bastion
```
crwr45, line 26: Is there a reason not to add the check to the previous play as well?

crwr45 (changes requested):

deploy_day2_workers.yml
nocturnalastro: Because hosts effectively does that when anyway if there are no hosts defined for it the play will be skipped.

PR #5, opened 2021-09-30
Linting fixes

crwr45:
Suggest reviewing commit-wise as changes are grouped together.


roles/get_image_hash/tasks/main.yml
```diff
@@ -1,5 +1,7 @@
 - name: check skopeo is installed
-  shell: /usr/bin/skopeo --version
+  package:
```
nocturnalastro, line 2: This will require root on the bastion, which we don't want to require.

nocturnalastro (approved):

mkochanowski (approved):

PR #8, opened 2021-10-07
Replace debug flag with verbosity level

nocturnalastro:
This uses the in build mechanisms of `debug` instead of recreating it with a flag


roles/boot_iso/tasks/kvm.yml
```diff
@@ -94,12 +94,12 @@
     status_code: [200, 201]
     validate_certs: no
     return_content: yes
+    verbosity: 1
```
crwr45, line 97: `uri` has no `verbosity` param

crwr45 (approved):

PR #12, opened 2021-10-19
Add ssh dir param to sample inventory

snecklifter:
The playbook errors out if we don't configure this.

### DIFF
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -94,6 +94,9 @@ all:
       - "{{ repo_root_path }}/ssh_public_key.pub"
       - ~/.ssh/id_rsa.pub
 
+    # Set the base directory to store ssh keys
+    ssh_key_dest_base_dir: /home/redhat
+
     # The retrieved cluster kubeconfig will be placed on the bastion host at the following location
     kubeconfig_dest_dir: /home/redhat/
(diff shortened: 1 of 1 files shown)


crwr45:
> > What's the error you get? Which version of Ansible have you got this on?
> > Looking at this I can't see why you have an error. https://github.com/redhat-partner-solutions/crucible/blob/main/roles/generate_ssh_key_pair/defaults/main.yml#L4
> 
> The error is that the directory doesn't exist, which on our server, it doesn't.

Ah, of course. 🤦  You are correct.

inventory.yml.sample
```diff
@@ -94,6 +94,8 @@ all:
       - "{{ repo_root_path }}/ssh_public_key.pub"
       - ~/.ssh/id_rsa.pub
 
+    ssh_key_dest_base_dir: /home/redhat
```
crwr45, line 97: Please can you add a short comment, similar to the one below (or extend that), to clarify under what circumstances a user might wish to modify this value, and to what.

snecklifter: Ok, I would have thought that was self-explanatory tbh.

nocturnalastro (approved):

PR #13, opened 2021-10-19
Fixes a few incorrect when conditions.

jtudelag:
Fixes a few incorrect when conditions.

### DIFF
--- a/roles/validate_inventory/tasks/network.yml
+++ b/roles/validate_inventory/tasks/network.yml
@@ -14,8 +14,7 @@
   changed_when: False
   loop: "{{ groups['services'] }}"
 
-- name: Ensure NTP server is available if not being set up
+- name: Ensure NTP server is reachable
   shell: # noqa 305
     cmd: "ping -c 1 -W 2 {{ ntp_server }}"
   changed_when: False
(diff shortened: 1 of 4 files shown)


roles/validate_inventory/tasks/network.yml
```diff
@@ -18,4 +18,4 @@
   shell: # noqa 305
     cmd: "ping -c 1 -W 2 {{ ntp_server }}"
   changed_when: False
-  when: (setup_ntp_service | default(True)) != True
+  when: setup_ntp_service | default(True)
```
crwr45, line 21: This change is inverting the (admittedly confusing) logic which I don't think is right, but on consideration I think this `when` should be entirely removed. The `{{ ntp_server }}` referenced must always be set so pinging it is always a valid check.
The `setup_ntp_service` flag only indicates if the host is already running the NTP service, not if the host is pingable.

Please also change the task name to match.

crwr45:
@jtudelag Have you got time to revisit and update this one?

jtudelag:
Hey @crwr45 no...I am flooded with other stuff, lets see if I have some time this week, apologies.

PR #16, opened 2021-10-21
Fix Day 2 worker node network config

snecklifter:
day2_workers weren't included in their network config task. This
commmit fixes this.


snecklifter:
Redundant, larger PR incoming.

PR #33, opened 2021-11-09
Updating chrony.conf

arjuhe:
Made changes to the chrony.conf template to allow crucible to:

- set the listening address to `{{ ntp_server }}`.
- set a default listen address of 127.0.0.1
- allow from 127.0.0.1
- Cleaned up chrony.conf file
- Enables logging on the server.

### DIFF
--- a/roles/setup_ntp/templates/chrony.conf.j2
+++ b/roles/setup_ntp/templates/chrony.conf.j2
@@ -1,20 +1,23 @@
 # {{ ansible_managed }}
-stratumweight 0
 driftfile /var/lib/chrony/drift
-rtcsync
-makestep 10 3
+bindcmdaddress {{ ntp_server }}
 bindcmdaddress 127.0.0.1
 bindcmdaddress ::1
 keyfile /etc/chrony.keys
-noclientlog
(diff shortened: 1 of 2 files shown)


roles/setup_ntp/templates/chrony.conf.j2
```diff
@@ -1,20 +1,23 @@
...
-logchange 0.5
-logdir /var/log/chrony
-manual
 local stratum 10
+rtcsync
+makestep 1.0 3
+manual
+
+logdir /var/log/chrony
+log measurements statistics tracking
```
nocturnalastro, line 13: Do we want to add this for all setups?

crwr45 (approved):

roles/setup_ntp/templates/chrony.conf.j2
arjuhe: This only applies to the ntp_host. And yes it will automatically set the listener to to the same address that we set the ntp_host to be.

crwr45 (approved):

nocturnalastro (approved):

PR #36, opened 2021-11-10
Install podman where required

nocturnalastro:


roles/setup_http_store/tasks/main.yml
```diff
@@ -46,14 +50,14 @@
         volumes:
           - "{{ http_data_dir }}:/var/www/html:z"
       register: container_info
-  become: true
 
 - name: Setting facts about container
```
crwr45, line 54: This seems to be the only task in the file not in a `become: true` block, and there's no reason for it not to be IMO.

crwr45 (approved):

roles/setup_http_store/tasks/main.yml
```diff
@@ -1,39 +1,43 @@
...
+
     - name: Create httpd container
       containers.podman.podman_pod:
         name: "{{ http_store_pod_name }}"
         publish:
           - 80:8080
       register: pod_info
 
-    - debug: # noqa unnamed-task
+    - debug:  # noqa unnamed-task
```
arjuhe, line 40: Extra space here.

arjuhe (commented):
I am still not a fan of using a block as a way to wrap become.

crwr45 (approved):

arjuhe (approved):
+1

PR #37, opened 2021-11-11
Allow multiple VMs for multiple cluster on the same host

nocturnalastro:


roles/destroy_vms/tasks/destroy_vms.yml
```diff
@@ -7,18 +7,9 @@
...
-  register: vms_list
-
-- name: Get filtered vm list
-  set_fact:
-    vms_to_remove: "{{ [item] + vms_to_remove | default([]) }}"
-  loop: "{{ vms_list.list_vms }}"
-  when: item in hostvars and hostvars[item]['vendor'] | lower == 'kvm'
+  loop: "{{ creation_to_remove.files | map(attribute='path') | select('match', prefix_regex) | list }}"
+  vars:
+    prefix_regex: "^{{ vm_create_scripts }}/?({{ vms_to_remove | join('|') }})_setup_vm.sh$"
```
mkochanowski, line 12: I think we could take this opportunity to rename the `vm_create_scripts` variable, as it can be confusing. (Is this a list of all VM scripts? Is this a function?)
This var configures the _directory_ for VM _create scripts_, so maybe we could add a `_dir` or `_path` suffix to the variable name to make it more descriptive.

roles/create_vms/templates/network.xml.j2
```diff
@@ -1,6 +1,6 @@
 <network>
   <name>{{ network_name }}</name>
-  <uuid>4414ebcb-8bdd-4cfc-a467-4376f3b709e9</uuid>
+  <uuid>{{ net_uuid | default(99999999 | random | to_uuid) }}</uuid>
```
mkochanowski, line 3: Noice

roles/destroy_vms/tasks/destroy_vms.yml
arjuhe: I agree with @mkochanowski we should be using prefix's and suffix's in var names to keep better track of them.

roles/destroy_vms/tasks/main.yml
```diff
@@ -1,32 +1,44 @@
...
+      loop: "{{ vms_list.list_vms }}"
+      when: item in cluster_vm_names
+
+- name: VMS to remove
+  debug:
+    var: vms_to_remove
+    verbosity: 1
+
+- name: Clean VM images
+  import_tasks: clean_vm_images.yml
```
crwr45, line 40: Is `command: undefine` not sufficient? https://libvirt.org/manpages/virsh.html#undefine suggests that an undefine (with the options all turned on) is complete enough, and https://github.com/ansible-collections/community.libvirt/blob/7cc0b5ff881fc2be108df7158e68a99057c8c952/plugins/modules/virt.py#L385 implies that the `command: undefine` option will remove everything.

crwr45 (approved):

roles/destroy_vms/tasks/main.yml
crwr45: Poorly explained: if `command: undefine` does not, as its docstring implies, "wipe it from the face of the earth", then a `virsh undefine --remove-all-storage --nvram` would be preferable for achieving the same result with less code than custom plays for extracting and removing volumes.

nocturnalastro: You we're correct thanks for the spot :)

crwr45 (approved):

arjuhe (approved):

roles/destroy_vms/tasks/main.yml
```diff
@@ -1,33 +1,40 @@
...
-    - "{{ bridge_name }}"
-    - "{{ vm_bridge_interface }}_{{ bridge_name }}"
-  ignore_errors: yes
-  become: true
-  when: (SETUP_VM_BRIDGE | default(true)) | bool == true
+    - name: Get filtered vm list
+      set_fact:
+        vms_to_remove: "{{ (vms_to_remove | default([])) + [item] }}"
+      loop: "{{ vms_list.list_vms }}"
+      when: item in cluster_vm_names
```
mkochanowski, line 32: I'd be cool if we could add this information to the deployment plan, as to make the user aware what VMs will be removed once Crucible is run. This might be out of scope for this PR, though.

mkochanowski (approved):

PR #39, opened 2021-11-15
Release image digest and assisted installer versions config.

nocturnalastro:
* Fix the issue with release images having tags instead of digests which breaks disconnected installs
* By default only configure assisted installer to deploy the version of openshift which we are trying to deploy.

### DIFF
--- a/roles/get_image_hash/defaults/main.yml
+++ b/roles/get_image_hash/defaults/main.yml
@@ -3,19 +3,28 @@ controller_tag: "{{ ai_version }}"
 installer_agent_tag: "{{ ai_version }}"
 installer_tag: "{{ ai_version }}"
 
-images_to_get_hash_for:
-  release:
-    image: quay.io/openshift-release-dev/ocp-release
-    tag: "{{ release_tag | default(openshift_full_version + '-x86_64') }}"
+ocp_release_versions_map:
+  '4.6': 4.6.16
+  '4.7': 4.7.33
+  '4.8': 4.8.14
+  '4.9': 4.9.0
+
+get_release_images: true
+
+release_image_url: "quay.io/openshift-release-dev/ocp-release"
+
+non_release_images:
   controller:
-    image: quay.io/ocpmetal/assisted-installer-controller
-    tag: "{{ controller_tag }}"
+      image: quay.io/ocpmetal/assisted-installer-controller
+      tag: "{{ controller_tag }}"
   installer_agent:
     image: quay.io/ocpmetal/assisted-installer-agent
     tag: "{{ installer_agent_tag }}"
   installer:
     image: quay.io/ocpmetal/assisted-installer
     tag: "{{ installer_tag }}"
+
+images_to_get_hash_for: "{{ non_release_images | combine(release_images | default({})) }}"
 
 destination_hosts:
   - registry_host
--- a/roles/get_image_hash/defaults/main.yml
+++ b/roles/get_image_hash/defaults/main.yml
@@ -3,20 +3,23 @@ controller_tag: "{{ ai_version }}"
 installer_agent_tag: "{{ ai_version }}"
 installer_tag: "{{ ai_version }}"
 
-ocp_release_versions_map:
+ocp_release_versions_map_all:
   '4.6': 4.6.16
   '4.7': 4.7.33
   '4.8': 4.8.14
   '4.9': 4.9.0
 
 get_release_images: true
+get_all_release_versions: false
+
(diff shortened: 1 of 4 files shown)


roles/get_image_hash/defaults/main.yml
```diff
@@ -3,20 +3,32 @@ controller_tag: "{{ ai_version }}"
...
+  '4.9': 4.9.0
+
+get_release_images: true
+get_all_release_version: false
+
+ocp_release_versions_map: "{{ get_all_release_version | ternary(ocp_release_versions_map_all, {openshift_version: ocp_release_versions_map_all[openshift_version]}) }}"
+
+release_image_url: "quay.io/openshift-release-dev/ocp-release"
+
+non_release_images:
```
mkochanowski, line 19: What's the distinction between non release and release images?

roles/setup_assisted_installer/defaults/main.yml
```diff
@@ -9,18 +9,18 @@ assisted_service_image: "quay.io/ocpmetal/assisted-service:{{ assisted_service_t
 assisted_service_gui_image: "quay.io/ocpmetal/ocp-metal-ui:{{ assisted_service_gui_tag }}"
 assisted_postgres_image: quay.io/ocpmetal/postgresql-12-centos7
 
-assisted_service_openshift_versions:
+assisted_service_openshift_versions_defaults:
   "4.6":
     display_name: 4.6.16
-    release_image: quay.io/openshift-release-dev/ocp-release:4.6.16-x86_64
+    release_image: "quay.io/openshift-release-dev/ocp-release@{{ image_hashes['release_4.6'] | default('4.6.16-x86_64') }}"
```
mkochanowski, line 15: Is an image name that uses a default value still valid after `:` was replaced with `@`?

roles/get_image_hash/defaults/main.yml
```diff
@@ -3,20 +3,32 @@ controller_tag: "{{ ai_version }}"
...
-    image: quay.io/openshift-release-dev/ocp-release
-    tag: "{{ release_tag | default(openshift_full_version + '-x86_64') }}"
+ocp_release_versions_map_all:
+  '4.6': 4.6.16
+  '4.7': 4.7.33
+  '4.8': 4.8.14
+  '4.9': 4.9.0
+
+get_release_images: true
+get_all_release_version: false
```
mkochanowski, line 13: I like these variable names (especially the `get_` prefix, we don't use that enough)! :+1: 
Was the var supposed to be named `get_all_release_versions`, though?

roles/get_image_hash/defaults/main.yml
```diff
@@ -3,20 +3,32 @@ controller_tag: "{{ ai_version }}"
...
+ocp_release_versions_map: "{{ get_all_release_version | ternary(ocp_release_versions_map_all, {openshift_version: ocp_release_versions_map_all[openshift_version]}) }}"
+
+release_image_url: "quay.io/openshift-release-dev/ocp-release"
+
+non_release_images:
   controller:
-    image: quay.io/ocpmetal/assisted-installer-controller
-    tag: "{{ controller_tag }}"
+      image: quay.io/ocpmetal/assisted-installer-controller
+      tag: "{{ controller_tag }}"
```
mkochanowski, line 22: It seems this indentation is redundant.

mkochanowski (changes requested):

roles/setup_assisted_installer/defaults/main.yml
nocturnalastro: Good spot :)

roles/get_image_hash/defaults/main.yml
nocturnalastro: Yes thanks :)

roles/get_image_hash/defaults/main.yml
nocturnalastro: It was just to distinguish that aren't the release image. perhaps `assisted_installer_images` would be better as that explains what they are used for more accurately.

crwr45: So `release_image` is referring to an OCP release, while non_release_image is for any other image?

nocturnalastro: Yes so we could have multiple release images the other will be singular. I think the `assisted_installer_images` is a good name for the other subset and it people want to add other subsets... well is ansible...

arjuhe (changes requested):
Address the issues brought up previously and I'll rereview. (:

crwr45 (approved):

mkochanowski (approved):

arjuhe (approved):
+1

PR #42, opened 2021-11-18
Allow KVM nodes to be hosted on multiple vm_hosts

nocturnalastro:


roles/validate_inventory/tasks/vms.yml
```diff
@@ -0,0 +1,83 @@
...
+- name: Check there is no vm_host in services
+  assert:
+    that:
+      - ('vm_host' not in groups['services'])
+    fail_msg: >
+      The structure of the inventory has changed
+      please put vm_host into a group called vm_hosts
+      and add a vm_host entry to the KVM node to point
+      to the host of the vm.
+  when: (kvm_node_names is defined) and (kvm_node_names | length > 0)
```
crwr45, line 18: nice idea using the validation to inform of a change to the Inventory. I don't think this needs to be conditional, though. It's a sensible check even if no VMs are defined.

roles/validate_inventory/tasks/vms.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Check that a host is defined in the 'vm_hosts' group if needed
+  assert:
+    that:
+      - groups['vm_hosts'] is defined
+      - groups['vm_hosts'] | length > 0
+    quiet: true
+  when: (kvm_node_names is defined) and (kvm_node_names | length > 0)
+
+- name: Check vm_host exists
```
crwr45, line 28: "Check that each node will be on a valid host" or something like that.

roles/validate_inventory/tasks/vms.yml
```diff
@@ -0,0 +1,83 @@
...
+  loop: "{{ kvm_node_names | default([]) }}"
+
+- name: Check KVM BMC username password combinations
+  block:
+    - name: Get KVM BMC username password combinations
+      vars:
+        kvm_user_password_combinations: {} # Structure will be {'{{vm_host}}.{{bmc_user}}': [{{bmc_password}}, ...]}
+      set_fact:
+        kvm_user_password_combinations: "{{
+            kvm_user_password_combinations | combine({
```
crwr45, line 43: Damnit, Ansible.

crwr45 (changes requested):

roles/display_deployment_plan/tasks/main.yml
```diff
@@ -2,64 +2,27 @@
...
-          Are you sure you want to proceed with the deployment using this inventory configuration?
-
-          Enter a value [yes/NO]
+        prompt: "{{ lookup('template', 'plan.j2').strip('\n') }}"
       vars:
         # Every row consists of two columns. To ensure all items in the right column are aligned,
         # a minimum width of the left column is set in the following variable.
         left_column_width: 26
-        row_format_str: '%-{{ left_column_width }}s %s'
+        row_format_str: "%-{{ left_column_width }}s %s"
```
mkochanowski, line 13: Do we have a convention on strings style for Crucible? Out of habit, I've been using single quotes `'` to describe machine-strings, and double quotes `"` for strings intended for humans. 

Not sure if we've converged on a specific set of rules for that. 
Some discussion about the _machine-string/human-string_ convention can be found [here](https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python).

roles/display_deployment_plan/templates/plan.j2
```diff
@@ -0,0 +1,41 @@
...
+* Groups and hosts configuration
+
+  {% for group_name in inventory.groups_filtered %}
+  {{group_name}}:
+  {% for host_name in groups[group_name] %}
+  {% if hostvars[host_name]['vendor'] is defined and hostvars[host_name]['vendor'] == 'KVM'  %}
+  {{ hosts_row_format_str | format(host_name, ((hostvars[host_name]['ansible_host'] +' (' + (hostvars[host_name]['vm_host'] | default(messages.value_missing)) + ')' ) | default(messages.value_missing))) }}
+  {% else %}
+  {{ hosts_row_format_str | format(host_name, (hostvars[host_name]['ansible_host'] | default(messages.value_missing))) }}
+  {% endif -%}
```
mkochanowski, line 29: I think it'd be possible to simplify this a tiny bit. We could create a new variable for rendering `ansible_host`, and push string concatenations out of filter invocations.

First, I'd add a dedicated formatting string for all rows extended with extra information (e.g. to display the `vm_host` assignment):
```yaml
hosts_row_extended_format_str: '%-{{ (left_column_width - hosts_row_padding) }}s %s (%s)'
```
Notice that the string has the extra `(%s)` at the end to display the info in brackets, so we don't have to do additional string concatenations anymore. 

Next, the expression for rendering `ansible_host` is being repeated here, and it gives us a chance to turn it into a dedicated variable. 

I think that the `if` condition can be simplified too. Removing the check for the `KVM` vendor will allow the user to catch errors with an invalid inventory config, e.g. baremetal nodes with assigned `vm_host`.

An updated snippet would look like the following:
```yaml
{%- set host_ansible_host = hostvars[host_name]['ansible_host'] | default(messages.value_missing) -%}
{% if hostvars[host_name]['vm_host'] is defined %}
{{ hosts_row_extended_format_str | format(host_name, host_ansible_host, hostvars[host_name]['vm_host']) }}
{% else %}
{{ hosts_row_format_str | format(host_name, host_ansible_host) }}
{% endif -%}
```

roles/process_kvm_nodes/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+vm_params:
+  master:
+    memory: "{{ vm_spec_master_memory | default(16384) }}"
+    vcpu: "{{ vm_spec_master_vcpu | default(6) }}"
+    disk_size: "{{ vm_spec_master_disk | default(30) }}"
+  worker:
+    memory: "{{ vm_spec_worker_memory | default(30000) }}"
+    vcpu: "{{ vm_spec_worker_vcpu | default(8) }}"
+    disk_size: "{{ vm_spec_worker_disk | default(50) }}"
```
mkochanowski, line 9: I'm not convinced it's best practice to have the default values embedded right inside the variable render expressions. 
In order to find a default value, a developer is forced to dig through all the `| default(number)` expressions, just to find where that value is defined. It's also easier to mistype the default value and let this go undetected.

I'd suggest separating the default data from logic. What if we used an isomorphic struct based on ` vm_params` to store the default values, and then reference the struct from `vm_params`?

See an example below:
```yaml
vm_default_params:
  master:
    memory: 16384
    vcpu: 6
    disk_size: 30
  worker:
    memory: 30000
    vcpu: 8
    disk_size: 50

vm_params:
  master:
    memory: "{{ vm_spec_master_memory | default(vm_default_params.master.memory) }}"
```

roles/process_kvm_nodes/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+vm_params:
+  master:
+    memory: "{{ vm_spec_master_memory | default(16384) }}"
+    vcpu: "{{ vm_spec_master_vcpu | default(6) }}"
+    disk_size: "{{ vm_spec_master_disk | default(30) }}"
```
mkochanowski, line 5: It's a little bit inconsistent to have one thing named `vm_params.master.disk_size` and the other `vm_spec_master_disk`. 
Maybe we could change `vm_spec_master_disk` to `vm_spec_master_disk_size`?

mkochanowski (changes requested):

roles/display_deployment_plan/tasks/main.yml
nocturnalastro: Oh sorry I think an auto-formatter did that. I thought I had turned it off. I think that machine/human makes sense although I don't think we use that consistently. I like the idea though. I'll try to track down what causes the change to see if its a ansible convention.

nocturnalastro: I changed it back and its not happening again. It must of happened before my auto formatter was turned off. I'll put it back to single quotes

roles/validate_inventory/tasks/vms.yml
```diff
@@ -0,0 +1,82 @@
+- name: Get all KVM Nodes
+  vars:
+    kvm_node_names: []
+  set_fact:
+    kvm_node_names: "{{ kvm_node_names + [item] }}"
+  when: hostvars[item]['vendor'] == 'KVM'
```
mkochanowski, line 6: The new role `process_kvm_nodes` uses case-insensitive conditional checks for `vendor`. Using the same style of checks would make the playbooks more consistent. Personally, I think that the case-insensitive ones are the way to go.
```yaml
when: hostvars[item]['vendor'] | lower == 'kvm'
```

mkochanowski (changes requested):

roles/validate_inventory/tasks/vms.yml
nocturnalastro: I agree I can add a commit and make all the vendor checks case insensitive. I don't think the ones in boot_iso or boot_disk are. What do you think?

mkochanowski: Sounds good

mkochanowski (approved):

roles/validate_inventory/tasks/vms.yml
```diff
@@ -79,4 +79,3 @@
         that:
           - (kvm_node_uuids | length) == (kvm_node_uuids | unique | length)
         fail_msg: "KVM node UUIDs must be unique otherwise they wont build or will collide durring discovery"
```
crwr45, line 81: `during`

crwr45 (approved):
One typo to fix, then all good

roles/validate_inventory/tasks/vms.yml
crwr45, line 81:
```suggestion
        fail_msg: "KVM node UUIDs must be unique otherwise they won't build or will collide during discovery"
```

PR #45, opened 2021-11-23
Fix various small issues.

nocturnalastro:


crwr45 (approved):

README.md
```diff
@@ -211,7 +211,7 @@ When doing multiple runs ensure you retain any authentication artefacts you need
 Existing tests can be run using
 
 ```bash
-ansible-playbook tests/run_tests.yml
+ANSIBLE_INVENTORY_UNPARSED_FAILED=false ansible-playbook tests/run_tests.yml
```
mkochanowski, line 214: Would adding the `unparsed_is_failed` setting to `tests/ansible.cfg` break the execution of our tests?
If we were to add this setting to the config file, we could then document that the official way of running tests is to use `ansible-playbook run_tests.yml` executed from the `tests/` directory, instead of requiring an environment variable to cover for an implementation detail.

inventory.yml.sample
```diff
@@ -182,7 +182,7 @@ all:
 
     vm_hosts:
       hosts:
-        vm_host: # Required for using "KVM" nodes, ignored if not.
+        my_vm_host: # Required for using "KVM" nodes, ignored if not.
```
mkochanowski, line 185: I think if we added a number to the sample `vm_host` name, it would give indication to the user that more than one `vm_host` is possible. One could say it shouldn't be needed because the group is named `vm_hosts` (plural), and to that I'd say: `bastions`.

We could name it `vm_host1`, as we already use (`super1`, `super2`, ...) and (`worker1`, `worker2`, ...) in our sample inventory.

mkochanowski (changes requested):

README.md
nocturnalastro: Good point I'll have a look but I think it should be fine.

inventory.yml.sample
```diff
@@ -52,7 +52,7 @@ all:
 
     # NTP Service
     # ntp_server is the address at which the NTP service is (or will be) available
-    ntp_server: ntp.example.lab
+    ntp_server:  10.40.0.100
```
mkochanowski, line 55: This space is redundant

mkochanowski (approved):

PR #48, opened 2021-11-29
Configure VM host network via nmstate

nocturnalastro:


crwr45 (approved):

arjuhe (commented):
I've been avoiding it, but I'll start this now.

roles/apply_nmstate/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+- name: Install nmstate
+  package:
+    name: nmstate
+    state: present
+
+- name: "Copy nmstate_yaml to {{ nmstate_config_path }}"
```
mkochanowski, line 6: Would `nmstate_rendered_yaml` be a better name for `nmstate_yaml`? As far as I understand, the variable holds a rendered template for _nmstate_.

roles/setup_vm_host_network/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+vm_bridge_ip: "{{ machine_network_cidr | ipaddr('next_usable') }}"
+vm_bridge_prefix: "{{ machine_network_cidr | ipaddr('prefix') }}"
+bridge_name: "{{ cluster_name }}-br"
```
mkochanowski, line 3: Missing `vm_` prefix here

roles/process_nmstate/defaults/main.yml
```diff
@@ -0,0 +1 @@
+target_name: "{{ inventory_hostname }}"
```
mkochanowski, line 1: With `{{ inventory_hostname }}` var assignments like these, I think it'd be a great practice to leave one or two sentences explaining why that's required for each scenario. This comment could make it a little easier for the reader to understand the execution flow.

mkochanowski (changes requested):

mkochanowski (approved):

arjuhe (approved):

PR #49, opened 2021-11-30
Update docs and inventory sample

nocturnalastro:


docs/inventory.md
```diff
@@ -210,22 +210,34 @@ That diagram gives the following excerpt from the inventory for the `bastion` an
       hosts:
         bastion:
           ansible_host: 192.168.10.5
+          ansible_connection: local # if your are not running crucible from the bastion then remove this line
```
arjuhe, line 213: I don't think the comment is needed here. It's not going to stop the PR but ansible_connection is basic ansible inventory stuff. Ultimately it makes the inventory messy and cluttered.

roles/setup_assisted_installer/defaults/main.yml
```diff
@@ -9,6 +9,7 @@ assisted_service_image: "quay.io/ocpmetal/assisted-service:{{ assisted_service_t
 assisted_service_gui_image: "quay.io/ocpmetal/ocp-metal-ui:{{ assisted_service_gui_tag }}"
 assisted_postgres_image: quay.io/ocpmetal/postgresql-12-centos7
 
+# Don't forget to update the sample inventory
```
arjuhe, line 12: I am thinking that all the comments in the inventory should get removed and replaced with a step by step guide to create the inventory from scratch.

roles/setup_sushy_tools/tasks/main.yml
```diff
@@ -68,7 +68,7 @@
             cert_file_prefix: sushy_tools
             cert_path: "{{ sushy_cert_dir }}/sushy_tools_{{ inventory_hostname }}.crt"
             privatekey_path: "{{ sushy_cert_dir }}/sushy_tools.key"
-            cert_common_name: "{{ ansible_fqdn }}"
+            cert_common_name: "{{ sushy_fqdn }}"
```
arjuhe, line 71: This file and roles/validate_inventory/tasks/vms.yml look like they were part of PR46. Did you intend for them to be here?

arjuhe (commented):
Take a look at roles/setup_sushy_tools/tasks/main.yml. It looks like it was part of PR46. Lastly, what is going on with roles/validate_inventory/tasks/vms.yml. Is this a new bug in the linter?

nocturnalastro:
> Take a look at roles/setup_sushy_tools/tasks/main.yml. It looks like it was part of PR46. Lastly, what is going on with roles/validate_inventory/tasks/vms.yml. Is this a new bug in the linter?

- It probably should have been PR46. i noticed it didn't exist while I was documenting stuff so I just added it in here as its only 2 lines it - it has is own commit though. 

- The linter thinks its logging the password but it isn't.

roles/setup_assisted_installer/defaults/main.yml
nocturnalastro: I'm not sure. I like the comments in the inventory as they are in the place of use if that makes sense. 
We can discuss it with the rest of the team.

roles/setup_sushy_tools/tasks/main.yml
nocturnalastro: It probably should have been PR46. i noticed it didn't exist while I was documenting stuff so I just added it in here as its only 2 lines it - it has is own commit though.

roles/setup_assisted_installer/defaults/main.yml
mkochanowski: I think leaving comments in the inventory is fine, although repeating the comment may be unnecessary here.

As soon as we wrap Crucible with something to simplify the UX of the interface (say a bash script or a python script) to get rid of the ansible-playbook command requiring multiple params, Ansible then becomes an implementation detail, something that is run in the background to do the heavy lifting throughout the deployment process.

Making things easier for people who don't have a lot of experience with Ansible by adding a sentence or two to the docs doesn't cost much, and facilitates transforming the goal from  "I want an OpenShift cluster deployed using Ansible" to "I want an OpenShift cluster deployed", which is an ultimately better goal. At least in my view :P

What I'd suggest is to remove this comment from the snippet as it's repeated multiple times, and instead, leave a note above the snippet explaining when and what `ansible_connection: local` is used for.

docs/inventory.md
mkochanowski: I think leaving comments in the inventory is fine, although repeating the comment may be unnecessary here.

As soon as we wrap Crucible with something to simplify the UX of the interface (say a bash script or a python script) to get rid of the ansible-playbook command requiring multiple params, Ansible then becomes an implementation detail, something that is run in the background to do the heavy lifting throughout the deployment process.

Making things easier for people who don't have a lot of experience with Ansible by adding a sentence or two to the docs doesn't cost much, and facilitates transforming the goal from  "I want an OpenShift cluster deployed using Ansible" to "I want an OpenShift cluster deployed", which is an ultimately better goal. At least in my view :P

What I'd suggest is to remove this comment from the snippet as it's repeated multiple times, and instead, leave a note above the snippet explaining when and what `ansible_connection: local` is used for.

arjuhe (approved):
None of the documentation updates block this PR for me. On the other hand we need to get the sushy tools fix merged.

mkochanowski (approved):

PR #50, opened 2021-11-30
Migrated a play. (CTONET-1053)

arjuhe:
Migrated 3 discreet populate_image_hash plays into 1 central playbook.

### DIFF
--- a/playbooks/deploy_registry.yml
+++ b/playbooks/deploy_registry.yml
@@ -1,16 +1,11 @@
 ---
-- name: Play to populate image_hashes for relevant images
-  hosts: localhost
+- import_playbook: ./populate_image_hashes.yml
   vars:
     destination_hosts:
       - registry_host
-  pre_tasks:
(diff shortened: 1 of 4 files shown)


playbooks/deploy_registry.yml
```diff
@@ -1,16 +1,11 @@
...
       - registry_host
-  pre_tasks:
-    - name: pre-compute need to get hashes
-      set_fact:
-        run_get_hash: "{{ image_hashes | default({}) | length == 0  }}"
-  roles:
-    - role: get_image_hash
-      when: run_get_hash
+    pre_compute: True
+  when: run_get_hash
```
nocturnalastro, line 7: does the when get inherited properly here? 

I don't thin this is the most readable way to do this. Perhaps we should move the when into playbook, we can then set the default for `run_get_hash: True` and use the `pre_compute` flag to figure out if it needs updated. 

I also think the flag could do with a more descriptive name.

PR #60, opened 2021-12-09
Add examples for day 2 workers

snecklifter:
A couple of items are missing from the inventory sample for when
adding day 2 workers. This commit adds them.

### DIFF
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -40,6 +40,27 @@ There three possible groups of nodes are `masters`, `workers` and `day2_workers`
 
 Day 2 nodes are added to an existing cluster. The reason why the installation of day 2 nodes is built into the main path of our automation is that for assisted installer day 2 nodes can be on a different L2 network which the main flow does not allow.
 
+Add a second iso name parameter to the inventory to avoid conflict with the original:
+
+```yaml
+# day2 workers require custom parameter
+day2_discovery_iso_name: "discovery/day2_discovery-image.iso"
+```
+
+Then add the stanza for day2 workers:
+
+```yaml
+day2_workers:
+  vars:
+    role: worker
+    vendor: HPE
+  hosts:
+    worker3: # Ensure this does not conflict with any existing workers
+      ansible_host: 10.60.0.106
+      bmc_address: 172.28.11.26
+      mac: 3C:FD:FE:78:AB:05
+```
+
 ### Network checks
 
 - All node `bmc_address`es are reachable


crwr45 (approved):

inventory.yml.sample
```diff
@@ -301,3 +304,21 @@ all:
 
               # # Uncomment to set an alternate installation disk device for the node
               # installation_disk_path: /dev/sdb
+
+        day2_workers:
+          vars:
+            role: worker
+            vendor: HPE # This example uses baremetal worker nodes
+          hosts:
+            worker1:
```
mkochanowski, line 313: `worker1` is already defined under `workers` in this file, which will lead to a silent name conflict. 

Unfortunately, the names have to be unique, as Ansible is flattening the hosts from all groups into a single list of hosts (`workers.worker1` and `day2_workers.worker1` become one `worker1` host).

Assume the nodes' vars are set to the following (values taken from the proposed sample inventory):

```
workers.worker1:
  mac: 3C:FD:FE:78:AB:03
  vendor: Dell
  # ...
    
day2_workers.worker1:
  mac: 3C:FD:FE:78:AB:05
  vendor: HPE
  # ...
```

The `worker1` hostvars will be:
```shell
$ ansible-inventory --host worker1 -i proposed-sample-inventory.yml | grep '"vendor"\|"mac"'
"mac": "3C:FD:FE:78:AB:05",
"vendor": "Dell",
```

As you can see, `worker1` is in disarray. The mac address is that of the `day2_workers.worker1` node, but the value of `vendor` was taken from `workers.worker1`.

mkochanowski (changes requested):
Thanks for the PR, all documentation is always very welcome!

However, could we move this change to the Inventory docs [1] instead of extending the sample inventory?
Currently, the sample inventory is intended to be a minimum inventory so that new users of Crucible have something easy to start with. More advanced configurations should be covered in the docs [1] instead.

There's a section about Day 2 Nodes at the top of the file, I think that could use the inventory snippet you've created :)

[1] Inventory docs: https://github.com/redhat-partner-solutions/crucible/blob/main/docs/inventory.md

inventory.yml.sample
snecklifter: Thanks for the review, have updated.

mkochanowski (approved):
Looks great! Thank you for your efforts.

nocturnalastro (approved):

PR #62, opened 2021-12-09
Fixes day2 worker playbook for disconnected installs

snecklifter:
Currently the day2 worker playbook is broken for disconnected installs
as the patching logic for mirrored registries does not exist. This
commit fixes this although its mostly a copy of the templates from
the main create cluster role so probably can be done a lot nicer.
Also, the CSR approval mechanism is still broken.

### DIFF
--- a/roles/create_day2_cluster/tasks/main.yml
+++ b/roles/create_day2_cluster/tasks/main.yml
@@ -48,3 +48,59 @@
   delegate_to: "{{ item }}"
   delegate_facts: yes
   loop: "{{ groups['day2_workers'] }}"
+
+#### patch discovery ignition on restricted network environments ###
+
+- name: Load patch for search registries
+  set_fact:
+    search_registries: "{{ lookup('template', 'patch-search-registries.j2') }}"
+  when: disconnected | bool == True
+
+- name: Load patch for discovery ignition
+  set_fact:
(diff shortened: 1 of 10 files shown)


nocturnalastro:
I agree I think we could refactor disconnected manifests into its own role and re-use it for the day2 process.

roles/create_day2_cluster/tasks/main.yml
```diff
@@ -48,3 +48,59 @@
...
+    body:
+      {
+        "config": "{{ patch_discovery_ignition | to_json(ensure_ascii=False) | string }}",
+      }
+  when: disconnected | bool == True
+  register: http_reply
+
+#### patch cluster install config on restricted network environments ###
+
+- name: Load patch for install config
```
nocturnalastro, line 81: Do you need the install config as well? 
have you tried with just the discovery patches?
I haven't looked at the day2 process to in depth but I'm wondering it will pick that up as machine configs from the cluster.

snecklifter: I'll need to revisit this as I can't recall and unable to test further at present. Should have a second shot in a few weeks.

crwr45 (approved):

nocturnalastro (approved):

PR #66, opened 2021-12-15
Add support for Single Node OpenShift deployments

mkochanowski:
From now on, end users can choose to deploy either a SNO or HA cluster.

SNO will be enabled automatically if and only if one master node and zero worker nodes are defined in the inventory.

Co-authored-by: Dylan Wong <dywong@redhat.com>
Signed-off-by: Marek Kochanowski <mkochanowski@redhat.com>

### DIFF
--- a/roles/create_cluster/defaults/main.yml
+++ b/roles/create_cluster/defaults/main.yml
@@ -20,8 +20,26 @@ HTTP_PROXY: ""
 HTTPS_PROXY: ""
 NO_PROXY: ""
 
-manifest_templates:
-  - 50-worker-nm-fix-ipv6.yml
-  - 50-worker-remove-ipi-leftovers.yml
-  - 02-fix-ingress-config.yml
-  - 01-master-node-scheduler.yml
+single_node_openshift_enabled: "{{ is_valid_single_node_openshift_config | default(false) }}"
+
+# Accepted values for "High Availability Mode"
+#
+# 1. 'None' (when `is_valid_single_node_openshift_config` evaluates to True)
+#     Requires one master node and no worker nodes -- SNO.
+# 2. 'Full' (otherwise)
+#     Requires at least three master nodes for High Availability.
+high_availability_mode: "{{ single_node_openshift_enabled | ternary('none', 'full') }}"
+
+manifest_templates_for_mode:
+  none:
+    - 50-worker-nm-fix-ipv6.yml
+    - 50-worker-remove-ipi-leftovers.yml
+    # - 02-fix-ingress-config.yml     # Applying this fix breaks the ingress controller on SNO.
+    # - 01-master-node-scheduler.yml  # The master node on SNO is automatically marked as schedulable.
+  full:
+    - 50-worker-nm-fix-ipv6.yml
+    - 50-worker-remove-ipi-leftovers.yml
+    - 02-fix-ingress-config.yml
+    - 01-master-node-scheduler.yml
+
+manifest_templates: "{{ manifest_templates_for_mode[high_availability_mode] }}"
--- a/roles/create_cluster/tasks/main.yml
+++ b/roles/create_cluster/tasks/main.yml
@@ -25,7 +25,7 @@
         "openshift_version": OPENSHIFT_VERSION,
         "base_dns_domain": BASE_DNS_DOMAIN,
         "cluster_network_cidr": CLUSTER_NETWORK_CIDR,
-        "cluster_network_host_prefix": 23,
+        "cluster_network_host_prefix": (CLUSTER_NETWORK_HOST_PREFIX | int),
         "service_network_cidr": SERVICE_NETWORK_CIDR,
         "pull_secret": (PULL_SECRET | to_json),
         "ssh_public_key": SSH_PUBLIC_KEY,
--- a/roles/create_cluster/tasks/main.yml
+++ b/roles/create_cluster/tasks/main.yml
@@ -1,7 +1,6 @@
 ---
 # tasks file for create_cluster
 
-# TODO: use the variable for cluster_network_host_prefix
 - name: Create cluster
   uri:
     url: "{{ URL_ASSISTED_INSTALLER_CLUSTERS }}"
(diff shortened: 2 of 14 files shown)


mkochanowski:
In this PR, all newly introduced vars were prefixed with the name of the role that they belong to.
I'd love to hear feedback from you on how you feel about the use of this convention.

roles/create_cluster/defaults/main.yml
```diff
@@ -20,8 +20,25 @@ HTTP_PROXY: ""
 HTTPS_PROXY: ""
 NO_PROXY: ""
 
-manifest_templates:
-  - 50-worker-nm-fix-ipv6.yml
-  - 50-worker-remove-ipi-leftovers.yml
-  - 02-fix-ingress-config.yml
-  - 01-master-node-scheduler.yml
+create_cluster__single_node_openshift_enabled: "{{ single_node_openshift_enabled | default(false) }}"
```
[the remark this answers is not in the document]
mkochanowski: Fair point! Let's discuss it with the team then.

One additional way would be to nest the vars under a key, like this:
```yaml
create_cluster:
  single_node_openshift_enabled: "{{ single_node_openshift_enabled | default(false) }}"
  high_availability_mode: "{{ create_cluster__single_node_openshift_enabled | ternary('None', 'Full') }}"
```
And then we could use `create_cluster.single_node_openshift_enabled` or even `create_cluster_vars.single_node_openshift_enabled`.

This might be a little bit more readable than `__`, as Ansible/jinja2 makes heavy use of the `.` notation.

roles/create_cluster/tasks/main.yml
```diff
@@ -11,24 +10,32 @@
...
+    #   type casted to a string. The Assisted Installer API requires this
+    #   input field to be an integer.
+    body: >-
+      {{ {
+        "name": CLUSTER_NAME,
+        "high_availability_mode": create_cluster__high_availability_mode,
+        "openshift_version": OPENSHIFT_VERSION,
+        "base_dns_domain": BASE_DNS_DOMAIN,
+        "cluster_network_cidr": CLUSTER_NETWORK_CIDR,
+        "cluster_network_host_prefix": (CLUSTER_NETWORK_HOST_PREFIX | int),
```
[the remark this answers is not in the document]
mkochanowski: Do you mean `CLUSTER_NETWORK_HOST_PREFIX`? If yes, then it's already defined in the playbook: https://github.com/redhat-partner-solutions/crucible/blob/main/playbooks/create_cluster.yml#L15.

The value was required but then wasn't consumed in any way :smiling_face_with_tear:

roles/create_cluster/defaults/main.yml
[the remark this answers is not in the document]
crwr45: +1 for using a key and then dot notation. I think the `_vars` is slightly too verbose for my taste.

roles/create_cluster/defaults/main.yml
```diff
@@ -20,8 +20,25 @@ HTTP_PROXY: ""
...
+
+# Accepted values for "High Availability Mode"
+# 1. 'None' (when `single_node_openshift_enabled` is True)
+#     Requires one master node and no worker nodes -- SNO.
+# 2. 'Full' (default)
+#     Requires at least three master nodes for High Availability.
+create_cluster__high_availability_mode: "{{ create_cluster__single_node_openshift_enabled | ternary('None', 'Full') }}"
```
crwr45, line 30: The Title case strings are counter-intuitive given (almost?) all the others are lowercase, or case insensitive. Make these lowercase and use a `| title` filter in the create cluster payload creation.

roles/create_cluster/defaults/main.yml
```diff
@@ -20,8 +20,25 @@ HTTP_PROXY: ""
...
+
+manifest_templates_for_mode:
+  'None':
```
crwr45, line 33: These keys would then also be lowercase and not need the `''`

crwr45 (changes requested):

roles/create_cluster/defaults/main.yml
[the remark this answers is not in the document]
arjuhe: I am alright with a single `_` for var names. `__` is tough to see and we should avoid it. Lastly Python and Jinja do strange things with dashes and fail to expand sometimes.

roles/create_cluster/defaults/main.yml
mkochanowski: Addressed

roles/create_cluster/defaults/main.yml
mkochanowski: Addressed

roles/create_cluster/defaults/main.yml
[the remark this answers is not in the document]
mkochanowski: I think it'd be best to move this conversation about style to a separate discussion. I've reverted the var prefixing change, all variables should now match the naming convention we've been using so far.

roles/validate_inventory/tasks/cluster.yml
```diff
@@ -5,11 +5,22 @@
...
+  when: not is_valid_single_node_openshift_config
+
+- name: Assert API and Ingress VIPs are set correctly (SNO)
+  assert:
+    that:
+      - api_vip == hostvars[sno_hostname]['ansible_host']
+      - ingress_vip == hostvars[sno_hostname]['ansible_host']
+    fail_msg: "For SNO deployments, API and Ingress VIPs need to match the IP address (ansible_host) of the master node."
+  vars:
+    sno_hostname: "{{ groups['masters'][0] | default('super1') }}"
```
nocturnalastro, line 22: At what point is the `when` evaluated? if it is before `vars` then I don't think you need a default here.

mkochanowski: You're right. Now that `is_valid_single_node_openshift_config` uses `groups['masters']` internally, the default here is redundant. I'll double-check that though.

nocturnalastro (approved):

roles/validate_inventory/tasks/cluster.yml
mkochanowski: I've tested the removal of the default (it works) but forgot to commit the change. Fixed!

arjuhe (approved):
LGTM

crwr45 (approved):

nocturnalastro (approved):

PR #69, opened 2022-01-07
Remove playbook vars, Simplify version setting

nocturnalastro:


inventory.yml.sample
```diff
@@ -12,8 +12,8 @@ all:
     cluster_name: clustername
     base_dns_domain: example.lab
 
-    # OpenShift version (4.6.16, 4.7.33, 4.8.14 or 4.9.0)
-    openshift_full_version: 4.6.16
+    # Openshift version (4.6, 4.7, 4.8)
```
crwr45, line 15: or 4.9

inventory.yml.sample
```diff
@@ -122,8 +122,13 @@ all:
...
     mirror_certificate: "{{ lookup('file', local_mirror_certificate_path) }}"
 
-    openshift_version: "{{ openshift_full_version.split('.')[:2] | join('.') }}"
+    openshift_full_version_mapping:
+      '4.6': 4.6.16
+      '4.7': 4.7.33
+      '4.8': 4.8.14
+      '4.9': 4.9.0
 
+    openshift_full_version: "{{ openshift_full_version_mapping[openshift_version | string] }}"
```
crwr45, line 131: How will this interact with an older inventory? If someone uses an old inventory, the new `openshift_full_version_mapping` which is used in `get_image_hash` will not exist. Add a helpful error message for this case, to ease the transition for people with an existing inventory they need to update.

crwr45 (changes requested):

inventory.yml.sample
```diff
@@ -122,8 +122,13 @@ all:
       }}"
     mirror_certificate: "{{ lookup('file', local_mirror_certificate_path) }}"
 
-    openshift_version: "{{ openshift_full_version.split('.')[:2] | join('.') }}"
+    openshift_full_version_mapping:
+      '4.6': 4.6.16
+      '4.7': 4.7.33
+      '4.8': 4.8.14
+      '4.9': 4.9.0
```
mkochanowski, line 129: I'm not sure if the mapping should be defined in the inventory. 

On the one hand, having it in the inventory makes it easier for the user to guess correctly what `openshift_version` will be expanded to.

On the other hand, it requires the user to manually update the inventory to have the mapping in line with `supported_ocp_versions` from `validate_inventory` after every new release is made. If the user still has to manually update the versions (just in a different place), what's the benefit?

Would it be possible to have the mapping defined somewhere in the roles' defaults instead? Are there any drawbacks to doing that?

nocturnalastro: Good spot. I forgot to revert this change I was playing around with moving defaults like this and `supported_ocp_versions` into a separate inventory. I'll move it into role defaults.

roles/validate_inventory/defaults/main.yml
```diff
@@ -20,8 +20,11 @@ supported_vendor_values:
...
-  - 4.8.14
-  - 4.9.0
+# This way a simple find and replace or an override of openshift_full_version_mapping will update the supported versions
+openshift_full_version_mapping:
+  '4.6': 4.6.16
+  '4.7': 4.7.33
+  '4.8': 4.8.14
+  '4.9': 4.9.0
+
+supported_ocp_versions: "{{ openshift_full_version_mapping.values() | list }}"
```
mkochanowski, line 30: We should assume that there could be more than one supported _patch_ version for every _minor_ version e.g. 
```
supported_ocp_versions: ['4.9.0', '4.9.10']
```

I'd suggest making `supported_ocp_versions` a base list of supported versions (it can be empty for now) which would additionally get extended with full version strings defined in `openshift_full_version_mapping`.

mkochanowski (changes requested):

inventory.yml.sample
```diff
@@ -12,8 +12,8 @@ all:
     cluster_name: clustername
     base_dns_domain: example.lab
 
-    # OpenShift version (4.6.16, 4.7.33, 4.8.14 or 4.9.0)
-    openshift_full_version: 4.6.16
+    # Openshift version (4.6, 4.7, 4.8 or 4.9)
+    openshift_version: 4.8
```
mkochanowski, line 16: Would it be possible to make `openshift_version` support both `x.y` and `x.y.z` patterns? 

**Example:**
- User provides a `x.y` version. The `openshift_full_version_mapping` is used to lookup the supported `x.y.z` version.
- User provides a `x.y.z` version. That version is used as `openshift_full_version`.

I think this could significantly simplify our input.

With the currently proposed implementation, do I have to override `openshift_full_version` in order to get a specific OpenShift version? 
If so, what would happen if I set the vars in my inventory to:
```yaml
openshift_version: 4.8
openshift_full_version: 4.9.0
```

nocturnalastro: Yes you would have to update the mapping. 
Err in terms of mismatched versions that probably wouldn't end well. We add a validation to check that doesn't happen

crwr45 (changes requested):
Lets remove the version setting part from this PR, its part of a larger issue for a later change.

nocturnalastro:
Need striping out and rebase  so I'm gonna close for now

PR #71, opened 2022-01-13
add boot flag so VM is restarted if host is rebooted

novacain1:
VMs ideally should have the autostart flag set, so if the physical server host itself is rebooted, the VMs are restarted without user intervention.

Feel free of course to edit as you see fit...


nocturnalastro (approved):
That's a good idea.

mkochanowski (approved):

arjuhe (changes requested):
This requires an enable disable var in defaults.

crwr45 (approved):

PR #72, opened 2022-01-21
KVM Role Consolidation.

arjuhe:
- merged all vm roles into one kvm role.
- collected all kvm playbooks into a kvm folder.
- Removed duplicate ansible from the various vm roles.
- added a task to validate vm_host can run vm's


arjuhe:
I am still working on this but wanted to get it up here so you guys can start looking at it.

playbooks/kvm/autostart.yml
```diff
@@ -0,0 +1,15 @@
...
+  import_playbook: process.yml
+  
+- name: Process Autostart on VMs
+  hosts: vm_hosts
+  tasks:
+    - import_role:
+        name: kvm
+        tasks_from: autostart
+  vars:
+    autostart: False
```
nocturnalastro, line 14: You wont be able to override this var in the inventory which I assume would be beneficial

playbooks/kvm/create.yml
```diff
@@ -0,0 +1,35 @@
...
+      import_role:
+        name: kvm
+        tasks_from: destroy
+      when: SETUP_VMS | bool
+      tags:
+        - destroy_vms
+    - name: Setup VM Host Network
+      import_role:
+        name: kvm
+        tasks_from: setup_up_host_network.yml
```
crwr45, line 25: This is the only `tasks_from` entry with the `.yml` in this file. Many other files include the `.yml`. Please make this standard across all uses.

I slightly prefer having the `.yml` as it makes it clear that it's loading a specific file, not doing any black magic under the hood akin to tags.

roles/kvm/tasks/destroy_vms.yml
```diff
@@ -0,0 +1,28 @@
...
+- name: "Remove file in {{ vm_create_scripts_dir }} dir"
+  file:
+    path: "{{ item }}"
+    state: absent
+  loop: "{{ creation_to_remove.files | map(attribute='path') | select('match', prefix_regex) | list }}"
+  vars:
+    prefix_regex: "^{{ vm_create_scripts_dir }}/?({{ vms_to_remove | join('|') }})_setup_vm.sh$"
+
+- name: Destroy and Undefine
+  import_tasks: destroy_block.yml
```
nocturnalastro, line 16: please put VM in the file and task name :)

roles/kvm/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- fail:
+    msg: |
+        Incorrect role inclusion.
+        This is intended to be run from the import_role module. Please see
+        https://docs.ansible.com/ansible/2.9/modules/import_role_module.html
```
nocturnalastro, line 7: Is there docs explaining the entry points?

roles/kvm/templates/create_vm.sh.j2
```diff
@@ -19,6 +20,7 @@ virt-install \
     --wait=-1 \
     --boot uefi \
     --events on_reboot=restart \
+    --autostart \
```
nocturnalastro, line 23: I feel like there should be an if here.

crwr45 (changes requested):
This looks like a nice change, converting a role from a function-like model into a library-like model of having a collection of related functionality together. I think this is potentially a good way to manage the complexity of large Ansible projects like this one. It's already showing some benefits here in terms of readability, reusability, and de-duplication.

However, this PR is almost impossible to properly review in its current state. Having all the file moves in a single commit, and including changes to the contents of those files in that commit makes it impossible to tell if any changes to the functionality have been made along the way, intended or not.
Please re-structure these changes into a series of smaller commits where as much as possible the file moves are in different commits to the changes to the file contents, or the consolidation happens one role at a time. This will make it far easier to review and to get this promising change in.

Please address the linter issues flagged here.

deploy_prerequisites.yml
```diff
@@ -16,6 +16,6 @@
 
 - import_playbook: playbooks/deploy_assisted_installer_onprem.yml
 
-- import_playbook: playbooks/create_vms.yml
+- import_playbook: playbooks/kvm/create.yml
```
nocturnalastro, line 19: Could you also factor out host network configuration as we need that to happen at the start of this play book as it will be required for some deployment scenarios.

nocturnalastro (changes requested):

PR #75, opened 2022-01-26
added become to apply_nmstate role.

arjuhe:


roles/apply_nmstate/tasks/main.yml
```diff
@@ -1,21 +1,22 @@
...
-    cmd: "nmstatectl apply --no-commit --timeout 120 {{ vm_nmstate_config_path }}"
-  async: 60
-  poll: 5
-
-- name: Commit changes
-  shell: # noqa: no-changed-when
-    cmd: "nmstatectl commit"
+---
+- name: Applying Nmstate
+  block:
```
mkochanowski, line 3: I'd like to see the `become: true` placed here (at the top of the block), similarly to how we do it in other parts of the code base, e.g. https://github.com/redhat-partner-solutions/crucible/blob/main/roles/create_vms/tasks/main.yml#L2-L13

roles/apply_nmstate/tasks/main.yml
```diff
@@ -1,21 +1,22 @@
...
+        mode: 0644
+      # No commit is done to revert the changes if they cause the host to be come unreachable
+    - name: Apply nmstate
+      shell: # noqa: no-changed-when
+        cmd: "nmstatectl apply --no-commit --timeout 120 {{ vm_nmstate_config_path }}"
+      async: 60
+      poll: 5
+    - name: Commit changes
+      shell: # noqa: no-changed-when
+        cmd: "nmstatectl commit"
```
mkochanowski, line 21: We should keep the newlines that used to separate these tasks. In my opinion, YAML squished in this way is quite difficult to read.

mkochanowski (changes requested):
Looks good. I proposed some formatting changes, details in the comments.

roles/apply_nmstate/tasks/main.yml
arjuhe: I think _blocks_ are hard to read as it is. 👍

mkochanowski (approved):

nocturnalastro (approved):

PR #77, opened 2022-01-28
Save kubeadmin credentials to a ansible vault

nocturnalastro:
Requires a new variable which `kubeadmin_vault_password_file_path` which defines the path to file containing the password for the vault

### DIFF
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -101,6 +101,8 @@ all:
     # The retrieved cluster kubeconfig will be placed on the bastion host at the following location
     kubeconfig_dest_dir: /home/redhat/
     kubeconfig_dest_filename: "{{ cluster_name }}-kubeconfig"
+    kubeadmin_dest_filename: "{{ cluster_name }}-kubeadmin.vault.yml"
+    # You can comment out the line below if you want the kubeadmin credentials to be stored in plain text
     kubeadmin_vault_password_file_path: "{{ repo_root_path }}/kubeadmin_vault_password_file"
 
     ############################
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -101,6 +101,7 @@ all:
     # The retrieved cluster kubeconfig will be placed on the bastion host at the following location
     kubeconfig_dest_dir: /home/redhat/
     kubeconfig_dest_filename: "{{ cluster_name }}-kubeconfig"
+    kubeadmin_vault_password_file_path: "{{ repo_root_path }}/kubeadmin_vault_password_file"
 
     ############################
     #    LOGIC: DO NOT TOUCH   #
--- a/post_install.yml
+++ b/post_install.yml
@@ -5,13 +5,12 @@
     secure: false
     ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ hostvars['assisted_installer']['host'] }}:{{ hostvars['assisted_installer']['port'] }}/api/assisted-install/v1"
     URL_ASSISTED_INSTALLER_CLUSTER: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/{{ cluster_id }}"
-    kube_filename: "{{ kubeconfig_dest_filename | default('kubeconfig') }}"
+    kube_filename: "{{ kubeconfig_dest_filename | default(cluster_name + '-kubeconfig') }}"
     dest_dir: "{{ kubeconfig_dest_dir | default(ansible_env.HOME) }}"
     kubeconfig_path: "{{ dest_dir }}/{{ kube_filename }}"
-
+    kubeadmin_vault_name: "{{ cluster_name }}-kubeadmin.vault.yml"
   environment:
     KUBECONFIG: "{{ kubeconfig_path }}"
-
   tasks:
     - name: Download kubeconfig
       get_url:
@@ -47,3 +46,12 @@
     - name: Login to add token to kubeconfig
       shell:
         cmd: "oc login -u {{ credentials.json.username }} -p '{{ credentials.json.password }}'"
+
+    - name: Save credentials to file
+      copy:
+        content: "{{ credentials.json | to_yaml }}"
+        dest: "{{ dest_dir }}/{{ kubeadmin_vault_name }}"
+
+    - name: Save credentials to vault
+      shell:
+        cmd: "ansible-vault encrypt --vault-password-file {{ kubeadmin_vault_password_file_path }} {{ dest_dir }}/{{ kubeadmin_vault_name }}"
--- a/post_install.yml
+++ b/post_install.yml
@@ -8,7 +8,7 @@
     kube_filename: "{{ kubeconfig_dest_filename | default(cluster_name + '-kubeconfig') }}"
     dest_dir: "{{ kubeconfig_dest_dir | default(ansible_env.HOME) }}"
     kubeconfig_path: "{{ dest_dir }}/{{ kube_filename }}"
-    kubeadmin_vault_name: "{{ cluster_name }}-kubeadmin.vault.yml"
+    kubeadmin_vault_name: "{{ kubeadmin_dest_filename | default(cluster_name +'-kubeadmin.vault.yml') }}"
   environment:
     KUBECONFIG: "{{ kubeconfig_path }}"
   tasks:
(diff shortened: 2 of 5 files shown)


roles/validate_inventory/tasks/required_vars.yml
```diff
@@ -4,3 +4,23 @@
...
+    - name: Stat kubeadmin_vault_password_file_path
+      stat:
+        path: "{{ kubeadmin_vault_password_file_path }}"
+      register: stat_kubeadmin_vault_password_file_path
+    - name: assert exists
+      assert:
+        that:
+          - stat_kubeadmin_vault_password_file_path.stat.exists
+        fail_msg: "kubeadmin_vault_password_file_path ({{ kubeadmin_vault_password_file_path }}) does not exist"
+      changed_when: False
```
mkochanowski, line 26: This can be simplified to a single assert task using a [path test](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-paths):
```yaml
- name: Check kubeadmin vault password file path is defined and the file exists
  assert:
    that:
      - kubeadmin_vault_password_file_path is defined
      - kubeadmin_vault_password_file_path is file
    fail_msg: "Kubeadmin Vault password must be stored in the location specified by the required variable 'kubeadmin_vault_password_file_path'."
  changed_when: False
```

post_install.yml
```diff
@@ -43,3 +44,13 @@
...
+
+    - name: Save credentials to file
+      ansible.builtin.copy:
+        content: "{{ credentials.json | to_yaml }}"
+        dest: "{{ dest_dir }}/{{ kubeadmin_vault_name }}"
+
+    - name: Save credentials to vault
+      # no_log: true
+      shell:
+        cmd: "ansible-vault encrypt --vault-password-file {{ kubeadmin_vault_password_file_path }} {{ dest_dir }}/{{ kubeadmin_vault_name }}"
```
mkochanowski, line 56: If we're using a FQCN for the `copy` task, I think it would make things more consistent if we added the `ansible.builtin.*` prefix to `shell` and all other relevant tasks too.

roles/validate_inventory/tasks/required_vars.yml
nocturnalastro: Good spot! I like the message as well :)

post_install.yml
nocturnalastro: Ah auto complete I'll remove the prefix on the copy :)

arjuhe (approved):
LGTM

snecklifter (approved):
LGTM

inventory.yml.sample
```diff
@@ -101,6 +101,8 @@ all:
     # The retrieved cluster kubeconfig will be placed on the bastion host at the following location
     kubeconfig_dest_dir: /home/redhat/
     kubeconfig_dest_filename: "{{ cluster_name }}-kubeconfig"
+    # You can comment out the line blow if you want the kubeadmin credentials to be stored in plain text
```
mkochanowski, line 104: fix typo: 'blow' --> 'below'

post_install.yml
```diff
@@ -5,13 +5,12 @@
     secure: false
     ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ hostvars['assisted_installer']['host'] }}:{{ hostvars['assisted_installer']['port'] }}/api/assisted-install/v1"
     URL_ASSISTED_INSTALLER_CLUSTER: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/{{ cluster_id }}"
-    kube_filename: "{{ kubeconfig_dest_filename | default('kubeconfig') }}"
+    kube_filename: "{{ kubeconfig_dest_filename | default(cluster_name + '-kubeconfig') }}"
     dest_dir: "{{ kubeconfig_dest_dir | default(ansible_env.HOME) }}"
     kubeconfig_path: "{{ dest_dir }}/{{ kube_filename }}"
-
+    kubeadmin_vault_name: "{{ cluster_name }}-kubeadmin.vault.yml"
```
mkochanowski, line 11: We have an inventory var for the kubeconfig filename (`kubeconfig_dest_filename`), but we don't have one for the kubeadmin vault -- its filename is hardcoded in `post_install.yml`.

It's a little bit inconsistent, but I won't be pushing back if you feel like adding a config var to the inventory is just not worth it.


```yaml
# The inventory's kubeconfig section could look like this
kubeconfig_dest_dir: /home/redhat/
kubeconfig_dest_filename: "{{ cluster_name }}-kubeconfig"
kubeadmin_dest_filename: "{{ cluster_name }}-kubeadmin.vault.yml"
```

roles/display_deployment_plan/templates/plan.j2
```diff
@@ -7,6 +7,8 @@ An OpenShift cluster is about to be deployed. Please double check the provided i
 
   {{ row_format_str | format('OpenShift version', inventory.openshift_full_version) }}
 
+  {{ row_format_str | format('Kube admin vault password file', inventory.kubeadmin_vault_password_file_path | default("UNDEFINED (This means the kubeadmin credentals will be stored in plain text)")) }}
```
mkochanowski, line 10: This won't work, unfortunately. You'll have to add  `kubeadmin_vault_password_file_path` to the deployment plan vars manually. The reason why you can't reference vars from the inventory directly is that Ansible will throw a templating error if an invalid inventory file is provided. Additionally, having it under the `inventory` key helps avoiding infinite loops caused by name conflicts.

You could start by extending the `messages` dict with your custom error message "UNDEFINED (This means...)"  here: https://github.com/redhat-partner-solutions/crucible/blob/main/roles/display_deployment_plan/tasks/main.yml#L27-L30

The second step is to add `kubeadmin_vault_password_file_path` to the `inventory` dict, referencing the new default error message. This can be added here: https://github.com/redhat-partner-solutions/crucible/blob/main/roles/display_deployment_plan/tasks/main.yml#L34-L48
```yaml
inventory:
    kubeadmin_vault_password_file_path: "{{ kubeadmin_vault_password_file_path | default(messages.kubeadmin_vault_file_path_missing) }}"
```

Then you should be able to print the row without the `default`:
```yaml
{{ row_format_str | format('Kube admin vault password file', inventory.kubeadmin_vault_password_file_path) }}
```

roles/display_deployment_plan/templates/plan.j2
```diff
@@ -25,6 +27,7 @@ An OpenShift cluster is about to be deployed. Please double check the provided i
   {{ row_format_str | format('Discovery ISO Name', inventory.discovery_iso_name) }}
   {{ row_format_str | format('Discovery ISO Download Path', inventory.discovery_iso_download_path) }}
 
+
```
mkochanowski, line 30: (possibly) redundant newline here

mkochanowski (changes requested):

post_install.yml
nocturnalastro: I think that makes sense We already did it for kubeconfig and it makes it easier for people to find the values

roles/display_deployment_plan/templates/plan.j2
nocturnalastro: Ah yes sorry that was me not reading the task properly.

inventory.yml.sample
```diff
@@ -101,6 +101,9 @@ all:
     # The retrieved cluster kubeconfig will be placed on the bastion host at the following location
     kubeconfig_dest_dir: /home/redhat/
     kubeconfig_dest_filename: "{{ cluster_name }}-kubeconfig"
+    kubeadmin_dest_filename: "{{ cluster_name }}-kubeadmin.vault.yml"
```
mkochanowski, line 104: `kubeadmin_dest_filename` is exposed in the inventory, but the var is not used anywhere. 

It should be referenced in `post_install.yml` in this var: https://github.com/redhat-partner-solutions/crucible/blob/c982fdbeaf7b051c16e08c84313876b205cf435b/post_install.yml#L11

mkochanowski (changes requested):

inventory.yml.sample
nocturnalastro: :facepalm:

nocturnalastro: Fixed now

mkochanowski (approved):
Pretty!

arjuhe (approved):
LGTM

mkochanowski:
I'll remove the "inventory breaking change" label, as the added variables are no longer mandatory and have defaults supplied.

PR #78, opened 2022-01-30
add Cleanup of /tmp/ssh_keys folder

halleelyasaf:
The play failed after running. The reason for the failure is that the file already exists with different permissions because of being created by a different user running crucible.
The file is also copied to the home directory of the user and doesn't use the /tmp directory anymore.
I would like to add a task to delete the files from the tmp folder.

### DIFF
--- a/roles/generate_ssh_key_pair/tasks/main.yml
+++ b/roles/generate_ssh_key_pair/tasks/main.yml
@@ -1,4 +1,9 @@
 ---
+- name: "Remove temporary ssh_keys folder {{ key_pair_dir }}"
+  file:
+    path: "{{ key_pair_dir }}"
+    state: absent
+
 - name: Make key dir
   file:
     path: "{{ key_pair_dir }}"


roles/generate_ssh_key_pair/tasks/main.yml
```diff
@@ -19,6 +19,11 @@
     - "{{ key_pair_dir }}/{{ private_key_name }}"
     - "{{ key_pair_dir }}/{{ public_key_name }}"
 
+- name: Remove ssh_keys folder from /tmp bastion
```
nocturnalastro, line 22: Due to ansible's policy of everything can be overwritten could you make the name a template so its something like 
```suggestion
- name: "Remove temporary ssh_keys folder {{ key_pair_dir }}"
```

halleelyasaf: Yes, You're right

mkochanowski (commented):
I guess that if the ssh keys have already been created and exist at runtime, then even with the proposed changes the play will still fail at `community.crypto.openssh_keypair` because the directory could not have been flushed (e.g. if the `fetch` task failed for some reason in a previous run).

My suggestion is to move the `file: state=absent` task to the top of the file, before the `file: state=directory` definition. This would ensure that the ssh keys can always be created, as it will be a clean directory every time.

halleelyasaf:
I think it's a mistake Because there is no module for move (mv) you are using the copy module and then you'll be left with junk after running.

halleelyasaf:
> I guess that if the ssh keys have already been created and exist at runtime, then even with the proposed changes the play will still fail at `community.crypto.openssh_keypair` because the directory could not have been flushed (e.g. if the `fetch` task failed for some reason in a previous run).
> 
> My suggestion is to move the `file: state=absent` task to the top of the file, before the `file: state=directory` definition. This would ensure that the ssh keys can always be created, as it's a clean directory every time.

I think because there is no module for move (mv) you are using the copy module and then you'll be left with junk after running.

halleelyasaf:
> I think Marek's point that moving the deletion of the folder before you re/create the folder, is a more robust way of ensuring there are no files to cause issues with the key pair gen task.
> 
> If task `Fetch SSH Key` fails then you wont reach the removal task where it currently is. If someone else was then to runs the play book the would hit the same error that this is solving.

I understand what he is saying and Marek definitely right.
But this does not solve the fact that there are files that you don't need to be left on the server.
how about adding one task before and one after?

mkochanowski:
> But this does not solve the fact that there are files that you don't need to be left on the server. How about adding one task before and one after?

I think that's a great point, although I would prefer if the scope of this PR was limited to addressing just the issue of ssh key creation failing (which moving the proposed `file:` task to the top of the file should fix). For the time being, I would not add a second clean-up task.

My intuition tells me that adding the capability of cleaning up artifacts created by Crucible would require applying a more general mechanism, including tracking the artifacts created. I suggest we explore different strategies for tracking and removing temporary artifacts in a separate PR, keeping in mind that the final mechanism would obviously have to work well with to the whole code base and all created artifacts.

halleelyasaf:
DONE

nocturnalastro (approved):
Can you squash the commits before we merge.

halleelyasaf:
> Can you squash the commits before we merge.

Sure

mkochanowski (approved):

arjuhe (approved):
👍

halleelyasaf:
> Can you squash the commits before we merge.

done

crwr45 (approved):

mkochanowski:
@halleelyasaf there's one tiny thing left, the branch needs to be up-to-date with `main`. Please rebase the branch when you get a chance.

halleelyasaf:
I tried to squash the merge commit into the add_cleenup and closed the PR by mistake

mkochanowski (approved):
Thank you for your contribution!

PR #80, opened 2022-02-03
Add ability to define partitions in ignition file.

hhamid93:
On branch main
Your branch is up to date with 'origin/main'.

Changes to be committed:
        modified:   roles/patch_host_config/tasks/main.yml
	new file:   roles/patch_host_config/templates/patch-worker-partition.j2

### DIFF
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -417,6 +417,17 @@ The basic network configuration of the inventory for the fully bare metal deploy
           bmc_address: 172.30.10.7
           # ...
 ```
+## Additional Partition Deployment
+
+For OCP 4.8+ deployments you can set partitions if required on the nodes. You do this by adding the snippet below to the node defination. Please ensure you provide the correct label and size(MiB) for the additional partitions you want to create. The device can either be the drive in which RHCOS image needs to be installed or it can be any additional drive on the node that requires partitioning. In the case that the device is equal to the host's `installation_disk_path` then a partition will be added defined by `disks_rhcos_root`. All additional partitions must be added under `extra_partitions` key as per the example below.
+
+```yaml
+disks:
+  - device: "{{ installation_disk_path }}"
+    extra_partitions:
+      partition_1: 1024
+      partition_2: 1024
+ ```     
 
 > **Note**: that the BMCs of the nodes in the cluster must be routable from the bastion host and the HTTP Store must be routable from the BMCs
 
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -294,7 +294,7 @@ all:
 
               # # Uncomment to set an alternate installation disk device for the node
               # installation_disk_path: /dev/sdb
-
+              
             worker2:
               ansible_host: 10.60.0.105
               bmc_address: 172.28.11.26
@@ -302,3 +302,5 @@ all:
 
               # # Uncomment to set an alternate installation disk device for the node
               # installation_disk_path: /dev/sdb
+             
+           
--- a/roles/patch_host_config/defaults/main.yml
+++ b/roles/patch_host_config/defaults/main.yml
@@ -1,4 +1,8 @@
 ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ hostvars['assisted_installer']['host'] }}:{{ hostvars['assisted_installer']['port'] }}/api/assisted-install/v1"
 URL_ASSISTED_INSTALLER_CLUSTER: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/{{ cluster_id }}"
-
 single_node_openshift_enabled: "{{ is_valid_single_node_openshift_config | default(false) }}"
+disks_rhcos_root: # Required rhcos_partion
+  label: root
+  number: 4
+  resize: true
+  sizeMiB: 102400
--- a/roles/patch_host_config/tasks/main.yml
+++ b/roles/patch_host_config/tasks/main.yml
@@ -95,3 +95,34 @@
           ]
         }
       register: http_reply
+
+- name: Load patch for worker partition
+  when: hostvars[host_name]['disks'] is defined
+  vars:
+    disks: "{{ hostvars[host_name]['disks'] }}"
+  block:
+    - name: Template patch partition
+      set_fact:
+        worker_partition: "{{ lookup('template', 'patch-worker-partition.j2') }}"
+
+    - debug: # noqa unnamed-task
+        var: worker_partition
+        verbosity: 1
+  
+  
+    - name: Patch worker partition
+      uri:
+        url: "{{ URL_ASSISTED_INSTALLER_CLUSTER }}/hosts/{{ host_id }}/ignition"
+        method: PATCH
+        status_code: [201]
+        return_content: True
+        body_format: json
+        body:
+          {
+        "config": "{{ worker_partition | to_json(ensure_ascii=False) | string }}",
+          }
+      register: http_reply
+   
+    - debug: # noqa unnamed-task
+        var: http_reply
+        verbosity: 1
--- a/roles/patch_host_config/templates/patch-worker-partition.j2
+++ b/roles/patch_host_config/templates/patch-worker-partition.j2
@@ -1,0 +1,28 @@
+#jinja2:trim_blocks: True, lstrip_blocks: True
+{
+  "ignition": {
+    "version": "3.2.0" 
+  },
+  "storage": {
+    "disks": [
+      {% for disk in disks %}
+      {
+        "device": "{{ disk.device }}",
+        "partitions": [ 
+          {% set partitions = ((disk.extra_partitions | default({})).items() | list) %}
+          {% if disk.device == (installation_disk_path | default(host.json.installation_disk_path)) %}
+          {{ (disks_rhcos_root | tojson) + (((partitions | length) > 0) | ternary(',','')) }}
+          {% endif %}
+          {% for label, size in partitions %}
+          {
+            "label": "{{ label | string }}",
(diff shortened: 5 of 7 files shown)


roles/patch_host_config/tasks/main.yml
```diff
@@ -93,3 +93,32 @@
           ]
         }
       register: http_reply
+
+- name: Load patch for worker partition
+  when: hostvars[host_name]['partitions'] is defined
+  block:
+    - name: Template patch partition
+      set_fact:
+        worker_partition: "{{ lookup('template', 'patch-worker-partition.j2') }}"
```
crwr45, line 102: The template is going to need to account for the configuration in `hostvars[host_name]['partitions']`, as it's currently hard-coded to 10x1GiB partitions, which is not going to work for most cases.

roles/patch_host_config/templates/patch-worker-partition.j2
```diff
@@ -0,0 +1,71 @@
+{
+	"ignition": {
+		"version": "3.2.0"
```
crwr45, line 3: We are going to need to come up with a way to bypass this functionality for older version of OCP, as ignition spec 3.2 is not supported on 4.6 (and possibly 4.7). I suggest adding something to the inventory validation step, and documenting the `partitions` field in the sample inventory.

crwr45 (changes requested):
This will be really useful to have baked in to Crucible rather than having to inject shell scripts to partition disks post-install.
There's a few changes to make to get it generic enough to be merged, though.

hhamid93:
Makes sense! 
Will try to get the template corrected so that user is able to decide on the number of partition if required or not.

Making this work with an older version is highly unlikely since 3.1.0 & below ignition spec does not have the "resize" parameter for making this work. 

Need to further discuss this with crucible team before making another commit.

mkochanowski:
Because the PR is still a work in progress, I converted it to a draft. 

Feel free to mark it as "ready for review" once it's finished. :)

docs/inventory.md
```diff
@@ -405,6 +405,20 @@ The basic network configuration of the inventory for the fully bare metal deploy
...
+## Additional Partition Deployment
+
+Uncomment the lines below in the inventory.yml to set additional partitions if required on the nodes. Please provide the correct label and size(MiB).
+Add all additional paritions required under extra_partitions (examples for label:size provided below).
+Additional partitions can only be created for OCP 4.8+ deployments
+
+```yaml
+# ...
+disks:
+  - device: "{{ installation_path }}" #DO NOT CHANGE
```
nocturnalastro, line 417: You might want to change that for instance if you have a second disk you wish to format

inventory.yml.sample
```diff
@@ -291,11 +291,32 @@ all:
...
               # # Uncomment to set an alternate installation disk device for the node
               # installation_disk_path: /dev/sdb
-
+              
+              # # Uncomment to set additional partitions if required on nodes. Please provide the correct label and size(MiB).
+              # # The device can either be the drive in which RHCOS image needs to be installed or it can be any additional drive that requires new partitions.
+              # # Add all additional paritions required under extra_partitions (examples for label:size provided below).
+              # # Additional partitions can only be created for OCP 4.8+ deployments
+              # disks:
+              # - device: "{{ installation_path }}" #See above section
```
mkochanowski, line 300: Should `installation_path` be `installation_disk_path` instead?

This is a snippet from **docs/inventory.md**:
```yaml
disks:
  - device: "{{ installation_disk_path }}"
    extra_partitions:
      partition_1: 1024
      partition_2: 1024
 ```

File: roles/patch_host_config/defaults/main.yml [hunk 5]
```diff
@@ -2,3 +2,9 @@ ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ hostv
 URL_ASSISTED_INSTALLER_CLUSTER: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/{{ cluster_id }}"
 
 single_node_openshift_enabled: "{{ is_valid_single_node_openshift_config | default(false) }}"
+
+disks_rhcos_root: # Required rhcos_partion
+  label: root
+  number: 4
```
mkochanowski, line 8: Does this mean that there will be 4 partitions?

inventory.yml.sample
hhamid93: Good Catch!
This has been corrected in the inventory.yml.sample
Thanks Marek!

roles/patch_host_config/templates/patch-worker-partition.j2
```diff
@@ -0,0 +1,28 @@
...
+    "version": "3.2.0" 
+  },
+  "storage": {
+    "disks": [
+      {% for disk in disks %}
+      {
+        "device": "{{ disk.device }}",
+        "partitions": [ 
+          {% set partitions = ((disk.extra_partitions | default({})).items() | list) %}
+          {% if disk.device == (installation_disk_path | default(host.json.installation_path)) %}
```
nocturnalastro, line 13: @hhamid93 Did you check the json returned from the host endpoint to check this is the correct path?

roles/patch_host_config/tasks/main.yml
```diff
@@ -95,3 +95,34 @@
...
+
+    - debug: # noqa unnamed-task
+        var: worker_partition
+        verbosity: 1
+  
+  
+    - name: Patch worker partition
+      uri:
+        url: "{{ URL_ASSISTED_INSTALLER_CLUSTER }}/hosts/{{host_id}}/ignition"
```
nocturnalastro, line 115: You need to add spaces around host id thats why theres a warning on 113.

hhamid93: Thanks ! 
This has been corrected

crwr45 (commented):
Looking great!

Needs the commits squashing, and since the behaviour is only v4.8 or higher it would be great to have a version check for this. I suggest adding a task to inventory validation to fail if the version is too low and the partition data is provided.

inventory.yml.sample
```diff
@@ -291,11 +291,32 @@ all:
...
+              # # Uncomment to set additional partitions if required on nodes. Please provide the correct label and size(MiB).
+              # # The device can either be the drive in which RHCOS image needs to be installed or it can be any additional drive that requires new partitions.
+              # # Add all additional paritions required under extra_partitions (examples for label:size provided below).
+              # # Additional partitions can only be created for OCP 4.8+ deployments
+              # disks:
+              # - device: "{{ installation_disk_path }}" #See above section
+              #   extra_partitions:
+              #     partition_1: 1024
+              #     partition_2: 102
+
```
arjuhe, line 304: This is an advanced feature and doesn't need to be in the sample inventory. The doc on how to use it should cover when, how, and where to use it.

inventory.yml.sample
```diff
@@ -291,11 +291,32 @@ all:
...
+              # # Uncomment to set additional partitions if required on nodes. Please provide the correct label and size(MiB).
+              # # The device can either be the drive in which RHCOS image needs to be installed or it can be any additional drive that requires new partitions.
+              # # Add all additional paritions required under extra_partitions (examples for label:size provided below).
+              # # Additional partitions can only be created for OCP 4.8+ deployments
+              # disks:
+              # - device: "{{ installation_disk_path }}" #See above section
+              #   extra_partitions:
+              #     partition_1: 1024
+              #     partition_2: 1024
+
```
arjuhe, line 322: This is redundant here.

roles/patch_host_config/templates/patch-worker-partition.j2
```diff
@@ -0,0 +1,28 @@
...
+          {% endif %}
+          {% for label, size in partitions %}
+          {
+            "label": "{{ label | string }}",
+            "sizeMiB": {{ size | int }}
+          }{{ ((loop.index == (partitions | length)) | ternary('',',')) }}
+          {% endfor %}
+        ],
+        "wipeTable": false
+      }{{ ((loop.index == (disks | length)) | ternary('',',')) }}
```
arjuhe, line 24: Should there be a new line between the curlies `}{{`?

arjuhe (changes requested):
Nothing major. Almost there.

roles/patch_host_config/templates/patch-worker-partition.j2
nocturnalastro: No we don't want a new line there as the comma that creates needs to be on end of the line loop is creating.

arjuhe: 👍

inventory.yml.sample
hhamid93: Change have been made. Removed from Sample inventory

inventory.yml.sample
hhamid93: change have been made.

arjuhe (approved):

arjuhe (changes requested):
> Looking great!
> 
> Needs the commits squashing, and since the behaviour is only v4.8 or higher it would be great to have a version check for this. I suggest adding a task to inventory validation to fail if the version is too low and the partition data is provided.

Please address the issue mentioned above.

hhamid93:
> > Looking great!
> > Needs the commits squashing, and since the behaviour is only v4.8 or higher it would be great to have a version check for this. I suggest adding a task to inventory validation to fail if the version is too low and the partition data is provided.
> 
> Please address the issue mentioned above.

hhamid93:
Validations have been added.

nocturnalastro (approved):
@arjuhe @crwr45 I think call comments have now been addressed

arjuhe (approved):
Congrats!!! +1

PR #90, opened 2022-02-28
Moves Ansible boot override mode to redfish_command

snecklifter:
With community.general 3.5.0, redfish_command supports overriding
the boot mode so lets use that rather than the uri module. This
should apply to other hardware vendors but I only have access to
ILO to test currently.


nocturnalastro (approved):
Matches what we do in dell https://github.com/redhat-partner-solutions/crucible/blob/main/roles/boot_iso/tasks/dell_redfish.yml#L21

PR #93, opened 2022-03-01
Add pxe boot

### LINKED ISSUE #92: Add the ability to deploy OCP using PXE boot
### Feature description

There are some vendors who don't fully support Redfish API. This prevents us from deploying OCP using Crucible in our lab.
Adding to Crucible the ability to deploy OCP clusters using PXE boot can help us to deploy OCP labs faster.

### Required statements

- [X] I acknowledge that Red Hat does not provide commercial support for the content of this repository.
- [X] I acknowledge that any assistance is offered purely on a best-effort basis, as resource permits.

halleelyasaf:
Adding the feature of installing crucible with PXE service on bare metal servers 
fix #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,16 @@ Requires the following to be installed on the deployment host:
 - [skopeo](https://github.com/containers/skopeo)
 - [podman](https://github.com/containers/podman/)
 - [kubectl + oc](https://docs.openshift.com/container-platform/4.9/cli_reference/openshift_cli/getting-started-cli.html)
+- [pyghmi](https://pypi.org/project/pyghmi/) #For PXE deployment
+- [ipmitool](https://github.com/ipmitool/ipmitool) #For PXE deployment
+
 
 **Important Note** The `openshift-clients` package is part of the (Red Hat OpenShift Container Platform Subscription
 )[https://access.redhat.com/downloads/content/290/]. The repo [must be activated on the bastion host](https://docs.openshift.com/container-platform/4.9/cli_reference/openshift_cli/getting-started-cli.html#cli-installing-cli-rpm_cli-developer-commands) before the dependency installation. It is used for the post-installation cluster validation steps.
 
 
 ```bash
-dnf install ansible python3-netaddr python3-jmespath skopeo podman openshift-clients
+dnf install ansible python3-netaddr skopeo podman openshift-clients ipmitool python3-pyghmi python3-jmespath
 ```
 
 There's also some required Ansible modules that can be installed with the following command:
@@ -165,6 +168,7 @@ Crucible can automatically set up the services required to deploy and run a clus
 - HTTP Store - Used to serve the Assisted Installer discovery ISO and allow it to be used as Virtual Media for nodes to boot from.
 - DNS - Optionally set up DNS records for the required cluster endpoints, and nodes. If not automatically set up then the existing configuration will be validated.
 - Assisted Installer - A pod running the Assisted Installer service, database store and UI. It will be configured for the target environment and is used by the cluster deployment playbook to coordinate the cluster deployment.
+- TFTP Host - A server that stores all the file mounted from the discovery image (required only for PXE deployments).
 
 While setup of each of these can be disabled if you wish to manually configure them, but it's highly recommended to use the automatic setup of all prerequisites.
 
@@ -205,6 +209,9 @@ The following are defaults for a full setup:
 - HTTP Store
   - A running pod containing the `httpd` service
   - The discovery image from Assisted Installer will be placed in and served from `/opt/http_store/data`
+
+- TFTP Host
+  - The discovery image will be mounted to this server and do the PXE boot with TFTP 
 
 
 ### Bastion
--- a/deploy_cluster.yml
+++ b/deploy_cluster.yml
@@ -6,6 +6,8 @@
 
 - import_playbook: playbooks/generate_discovery_iso.yml
 
+- import_playbook: playbooks/mount_discovery_iso_for_pxe.yml
+
 - import_playbook: playbooks/boot_iso.yml
   vars:
     boot_iso_hosts: masters,workers
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -218,6 +218,7 @@ Use the following vars to control setup of prerequisites:
 - `setup_dns_service`
 - `create_vms`
 - `setup_sushy_tools`
+- `setup_pxe_service`
 
 Note that if one or more of these services is pre-existing in your environment the inventory must still be configured with information needed to access those services, even when the service is not being set up by the playbooks.
 
@@ -429,6 +430,51 @@ disks:
       partition_2: 1024
  ```     
 
+## PXE Deployment
+You must have these services when using PXE deployment
+- `DHCP`
+- `DNS`
+- `PXE`
+```
+       masters:
+          vars:
+            role: master
+            vendor: pxe # this example is a PXE control plane
+          hosts:
+            super1:
+              ansible_host: 10.60.0.101
+              mac: "DE:AD:BE:EF:C0:2C"
+              bmc_address: "192.168.10.16"
+
+
+            super2:
+              ansible_host: 10.60.0.102
+              mac: "DE:AD:BE:EF:C0:2D"
+              bmc_address: "192.168.10.17"
+
+
+            super3:
+              ansible_host: 10.60.0.103
+              mac: "DE:AD:BE:EF:C0:2E"
+              bmc_address: "192.168.10.18"
+
+        workers:
+          vars:
+            role: worker
+          hosts:
+            worker1:
+              ansible_host: 10.60.0.104
+              bmc_address: 192.168.10.19
+              mac: 3c:fd:fe:b5:79:04
+              vendor: pxe
+            worker2:
+              ansible_host: 10.60.0.105
+              bmc_address: 192.168.10.20
+              mac: "DE:AD:BE:EF:C0:2F"
+              vendor: pxe
+              bmc_address: "nfvpe-21.oot.lab.eng.bos.redhat.com:8082"
+   
+```
 > **Note**: that the BMCs of the nodes in the cluster must be routable from the bastion host and the HTTP Store must be routable from the BMCs
 
 These two examples are not the only type of clusters that can be deployed using Crucible. A hybrid cluster can be created by mixing virtual and bare metal nodes.
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -45,6 +45,7 @@ all:
     # You will need to ensure alternatives are available for anything that will not be automatically set up
     setup_ntp_service: false
     setup_dns_service: false
+    setup_pxe_service: false
     setup_registry_service: false # Only required for a Restricted Network installation
     setup_http_store_service: true
     setup_assisted_installer: true # default is true you may wish to turn it off if multiple users are using the same instance.
@@ -172,6 +173,7 @@ all:
           # upstream_dns: 8.8.8.8 # an optional upstream dns server
           # The following are required for DHCP setup
           # use_dhcp: true
+          # use_pxe: false
           # dhcp_range_first: 10.60.0.101
           # dhcp_range_last:  10.60.0.105
           # prefix: 24
@@ -179,6 +181,10 @@ all:
 
         http_store:
           ansible_host: 10.40.0.100
+        
+        tftp_host
+          ansible_host: "{{ hostvars['dns_host']['ansible_host'] }}"
+          tftp_directory: /var/lib/tftpboot/ 
 
         ntp_host:
           ansible_host: 10.40.0.100
@@ -217,10 +223,11 @@ all:
       #     - The value set here will be used when creating VMs and must be unique within the network.
       #
       # - vendor
-      #     - One of "Dell", "HPE", "Lenovo", "SuperMicro", "KVM" as the supported BMC APIs.
+      #     - One of "Dell", "HPE", "Lenovo", "SuperMicro", "KVM", "PXE" as the supported BMC APIs.
       #     - "KVM" identifies a node as a VM to be created. If a "KVM" node is present,
       #       then a "vm_host" must be defined in the node and a host with that name must exist
-      #       inside the "vm_hosts" group.
+      #       inside the "vm_hosts" group. 
+      #     - "PXE" identifies a node as a baremetal that needs to boot from PXE.
       #
       # - bmc_address
       # - bmc_user
--- a/playbooks/README.md
+++ b/playbooks/README.md
@@ -10,6 +10,7 @@ and are nominally put in the order of usage with some exceptions. This may be us
 | `deploy_http_store.yml`                | Deploys HTTP server                                           | -                     |
 | `deploy_registry.yml`                  | Deploys container image registry                              | -                     |
 | `deploy_dns.yml`                       | Deploys DNS and DHCP                                          | -                     |
+| `deploy_tftp.yml`                      | Deploys tftp server                                          | -                     |
 | `deploy_assisted_installer_onprem.yml` | Deploys podman pod with assisted service containers           | -                     |
 | `create_vms.yml`                       | Destroys and creates VMs and bridge                           | -                     |
 | `deploy_sushy_tools.yml`               | Deploys sushy tools                                           | -                     |
@@ -18,6 +19,7 @@ and are nominally put in the order of usage with some exceptions. This may be us
 | `create_cluster.yml`                   | Creates cluster definition in assisted installer              | -                     |
 | `generate_discovery_iso.yml`           | Generate descovery iso                                        | `cluster_id`          |
 | `boot_iso.yml`                         | Reboots nodes to discovery iso                                | -                     |
+| `mount_discovery_iso_for_pxe.yml`      |  Mounts discovery iso, generates grub config and syncs files to the TFTP server                                 | -                     |
 | `install_cluster.yml`                  | Trigger install of discovered hosts                           | `cluster_id`          |
 | `monitor_hosts.yml`                    | Monitor hosts for discovery and to become ready               | `cluster_id`          |
 | `monitor_cluster.yml`                  | Monitor cluster for installation and initalisation            | `cluster_id`          |
--- a/playbooks/deploy_tftp.yml
+++ b/playbooks/deploy_tftp.yml
@@ -1,0 +1,7 @@
+- name: Setup TFTP
+  hosts: tftp_host
+  vars:
+    SETUP_TFTP_SERVICE: "{{ setup_pxe_service | default(False) }}"
+  roles:
+    - role: setup_tftp
+      when: SETUP_TFTP_SERVICE == True
--- a/playbooks/mount_discovery_iso_for_pxe.yml
+++ b/playbooks/mount_discovery_iso_for_pxe.yml
@@ -1,0 +1,6 @@
+- hosts: http_store
+  vars:
+    MOUNT_DISCOVERY_ISO: "{{ setup_pxe_service | default(False)}}"
+  roles:
+    - role: mount_discovery_iso_for_pxe
+      when: MOUNT_DISCOVERY_ISO == True
--- a/roles/insert_dns_records/tasks/configure_firewall.yml
+++ b/roles/insert_dns_records/tasks/configure_firewall.yml
@@ -20,3 +20,27 @@
     - internal
     - public
   when: use_dhcp == true
+
+- name: Open port in firewall for proxy DHCP
+  ansible.posix.firewalld:
+    port: "4011/udp"
+    permanent: yes
+    immediate: yes
+    state: enabled
+    zone: "{{ item }}"
+  loop:
+    - internal
+    - public
+  when: use_pxe == true
+
+- name: Open port in firewall for PXE
+  ansible.posix.firewalld:
+    port: "69/udp"
+    permanent: yes
+    immediate: yes
+    state: enabled
+    zone: "{{ item }}"
+  loop:
+    - internal
+    - public
+  when: use_pxe == true
--- a/roles/mount_discovery_iso_for_pxe/defaults/main.yml
+++ b/roles/mount_discovery_iso_for_pxe/defaults/main.yml
@@ -1,0 +1,42 @@
+---
+# defaults file for generate_discovery_iso
+tftp_root: "{{ hostvars['tftp_host']['tftp_directory'] | default('/var/lib/tftpboot/') }}"
+mount_images_directory: "/mnt/iso/images"
+mount_efiboot_directory: "/mnt/efiboot"
+pxe_directory: "pxe"
+rootfs_directory: "/opt/http_store/data/{{pxe_directory}}"
+mounted_iso_files:
+  - "{{ mount_images_directory }}/ignition.img"
+  - "{{ mount_images_directory }}/efiboot.img"
+  - "{{ mount_images_directory }}/pxeboot/vmlinuz"
+  - "{{ mount_images_directory }}/pxeboot/initrd.img"
+  - "{{ mount_efiboot_directory }}/EFI/redhat/grubx64.efi"
+  - "{{ mount_efiboot_directory }}/EFI/redhat/shimx64.efi"
+
+mounted_tftp_files: "http://{{ HTTPD_PXE_HOST }}/{{ pxe_directory }}"
+tftp_files_to_download:
+  - "{{ mounted_tftp_files }}/ignition.img"
+  - "{{ mounted_tftp_files }}/efiboot.img"
+  - "{{ mounted_tftp_files }}/vmlinuz"
+  - "{{ mounted_tftp_files }}/initrd.img"
+  - "{{ mounted_tftp_files }}/grubx64.efi"
+  - "{{ mounted_tftp_files }}/shimx64.efi"
+
+system_files_to_download:
+  - "{{ mounted_tftp_files }}/lpxelinux.0" 
+  - "{{ mounted_tftp_files }}/ldlinux.c32"  
+  - "{{ mounted_tftp_files }}/menu.c32"
+
+system_files_directory: "/usr/share/syslinux"
+system_files:
+  - "{{ system_files_directory }}/lpxelinux.0"
+  - "{{ system_files_directory }}/ldlinux.c32"
+  - "{{ system_files_directory }}/menu.c32"
+
+ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ ASSISTED_INSTALLER_HOST }}:{{ ASSISTED_INSTALLER_PORT }}/api/assisted-install/v1"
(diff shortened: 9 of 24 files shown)


roles/generate_ssh_key_pair/tasks/main.yml
```diff
@@ -19,6 +19,11 @@
     - "{{ key_pair_dir }}/{{ private_key_name }}"
     - "{{ key_pair_dir }}/{{ public_key_name }}"
 
+- name: Remove ssh_keys folder from /tmp bastion
```
nocturnalastro, line 22: Hasn't this already been merged?

halleelyasaf: Yes, it's weird that it's here

halleelyasaf: I removed the changes

nocturnalastro (changes requested):

roles/setup_pxe_httpd/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Setup httpd
```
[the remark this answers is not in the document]
halleelyasaf: like you said we discussed on that and we don't want to change stuff in the crucible itself.

docs/inventory.md
```diff
@@ -406,6 +409,51 @@ The basic network configuration of the inventory for the fully bare metal deploy
           # ...
 ```
 
+## PXE Deployment
+this services need to set as true when using PXE deployment
+- `DHCP`
```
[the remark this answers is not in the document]
halleelyasaf: It does not need to be enabled you can use different DHCP&DNS.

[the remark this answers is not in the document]
halleelyasaf: The definition is inaccurate I will change

roles/mount_discovery_iso_for_pxe/templates/grub.cfg.j2
```diff
@@ -0,0 +1,5 @@
+set timeout=1
+menuentry 'Install Red Hat Enterprise Linux CoreOS' --class fedora --class gnu-linux --class gnu --class os {
+	linuxefi vmlinuz random.trust_cpu=on ignition.firstboot ignition.platform.id=metal 'coreos.live.rootfs_url=http://{{HTTPD_PXE_HOST}}:{{HTTPD_PORT_FOR_PXE}}/rootfs.img'
```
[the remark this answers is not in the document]
halleelyasaf: this file is for the PXE, it tells the boot server where is the file that he needs to pull located.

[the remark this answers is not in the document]
halleelyasaf: it does not matter how many clusters you need to deploy.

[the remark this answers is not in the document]
nocturnalastro: Hmm OK thinking about it seems like the failure case is much more fundamental to the method than just the rootfs url.

roles/setup_pxe_httpd/tasks/main.yml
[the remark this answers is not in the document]
nocturnalastro: Looking at this though it doesn't really look like any changes are needed. Just dump files into a directory. Not sure its worth an extra host and setting up of another http server. 

Its defiantly worth trying just dumping the files into a directory on the webroot of the normal of the http sever.

playbooks/deploy_httpd_for_pxe.yml
```diff
@@ -0,0 +1,7 @@
+- name: Setup httpd for PXE
+  hosts: httpd_pxe_host
```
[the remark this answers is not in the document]
halleelyasaf: we will agree to delete that and use the HTTP-store, we will delete that.

playbooks/deploy_httpd_for_pxe.yml
```diff
@@ -0,0 +1,7 @@
...
+  vars:
+    SETUP_HTTPD_SERVICE: "{{setup_pxe_service | default(True)}}"
```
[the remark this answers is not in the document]
halleelyasaf: need to add space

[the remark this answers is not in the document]
halleelyasaf: add a check of if you using vendor=pxe check if the setup_pxe_service= true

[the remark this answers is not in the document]
halleelyasaf: remove that

playbooks/deploy_httpd_for_pxe.yml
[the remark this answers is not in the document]
halleelyasaf: remove that

playbooks/mount_discovery_iso_for_pxe.yml
```diff
@@ -0,0 +1,6 @@
+- hosts: bastion
+  vars:
+    MOUNT_DISCOVERY_ISO: "{{setup_pxe_service | default(True)}}"
```
arjuhe, line 3: This is making the assumption that setup_pxe is true if undefined. This should default to false.

playbooks/deploy_tftp.yml
```diff
@@ -0,0 +1,7 @@
+- name: Setup TFTP
+  hosts: tftp_host
+  vars:
+    SETUP_TFTP_SERVICE: "{{ setup_pxe_service | default(True) }}"
```
arjuhe, line 4: The default for setup_pxe_service shouldn't default to true.

deploy_cluster.yml
```diff
@@ -6,6 +6,8 @@
 
 - import_playbook: playbooks/generate_discovery_iso.yml
 
+- import_playbook: playbooks/mount_discovery_iso_for_pxe.yml
```
arjuhe, line 9: Please a when here to make it clear when this playbook gets run.

arjuhe (changes requested):
Just a few small changes

playbooks/mount_discovery_iso_for_pxe.yml
arjuhe: One more thing. Please put a space at the end of '{{' and the begining of '}}'

roles/mount_discovery_iso_for_pxe/templates/grub.cfg.j2
```diff
@@ -0,0 +1,5 @@
+set timeout=1
+menuentry 'Install Red Hat Enterprise Linux CoreOS' --class fedora --class gnu-linux --class gnu --class os {
+	linuxefi vmlinuz random.trust_cpu=on ignition.firstboot ignition.platform.id=metal 'coreos.live.rootfs_url=http://{{ HTTPD_PXE_HOST }}/{{ pxe_directory }}/rootfs.img'
```
natifridman, line 3: I think we should add the HTTP port from `HTTPD_PORT_FOR_PXE`

inventory.yml.sample
```diff
@@ -176,7 +177,14 @@ all:
 
         http_store:
           ansible_host: 10.40.0.100
+        
+        tftp_host
+          ansible_host: 10.40.0.100
+          tftp_directory: /var/lib/tftpboot/ 
 
+        httpd_pxe_host:
```
natifridman, line 185: Do we need this? I thought we use the `http_store`

roles/mount_discovery_iso_for_pxe/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  - "{{ system_files_directory }}/lpxelinux.0"
+  - "{{ system_files_directory }}/ldlinux.c32"
+  - "{{ system_files_directory }}/menu.c32"
+ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ ASSISTED_INSTALLER_HOST }}:{{ ASSISTED_INSTALLER_PORT }}/api/assisted-install/v1"
+URL_ASSISTED_INSTALLER_CLUSTERS_DOWNLOAD_IMAGE: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/{{ CLUSTER_ID }}/downloads/image"
+
+DOWNLOAD_DEST_FILE: "{{ discovery_iso_name }}"
+DOWNLOAD_DEST_PATH: "{{ iso_download_dest_path | default('/opt/http_store/data') }}"
+HTTPD_PXE_HOST: "{{ hostvars['httpd_pxe_host']['ansible_host'] }}"
+HTTPD_PORT_FOR_PXE: "{{ hostvars['httpd_pxe_host']['httpd_port_for_pxe'] }}"
```
natifridman, line 28: Do we need this? I thought we use the `http_store`

roles/insert_dns_records/tasks/configure_firewall.yml
```diff
@@ -20,3 +20,39 @@
...
+  ansible.posix.firewalld:
+    port: "69/udp"
+    permanent: yes
+    immediate: yes
+    state: enabled
+    zone: "{{ item }}"
+  loop:
+    - internal
+    - public
+  when: use_pxe == true
```
mkochanowski, line 58: It should be possible to merge all these tasks into one, iterating over a list of ports.
```yaml
pxe_required_open_ports:
  - "4011/udp"
  - "81/tcp"
  - "69/udp"
```

You'll probably need to use a nested loop for that: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#iterating-over-nested-lists

roles/setup_ipmi_tools/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: Install python3-pip
+      ansible.builtin.package:
+        name: python3-pip
+        state: present
+    
+    - name: Install python for the ipmi tool module
+      pip:
+        name:
+          - pyghmi
+  become: true
```
mkochanowski, line 18: Please move the `become` to the top of the block:
```yaml
- name: Setup IPMI
  become: true
  block:
    - name: Install IPMI tools
      # ...
```

roles/setup_tftp/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+        enabled: true
+
+    - name: Allow incoming tftp traffic
+      ansible.posix.firewalld:
+        zone: public
+        service: tftp
+        permanent: yes
+        state: enabled
+        immediate: yes
+  become: true
```
mkochanowski, line 23: Please move the `become` to the top of the block:
```yaml
- name: Setup TFTP
  become: true
  block:
    - name: Install TFTP
      # ...
```

roles/setup_tftp/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Setup TFTP
+  block:
+    - name: Install TFTP
+      ansible.builtin.package:
+        name: tftp-server
+        state: present
+
+
```
mkochanowski, line 9: Redundant newline here.

mkochanowski (changes requested):

roles/mount_discovery_iso_for_pxe/templates/grub.cfg.j2
halleelyasaf: I will add that after the fix of the HTTP port

roles/mount_discovery_iso_for_pxe/defaults/main.yml
halleelyasaf: I will delete the host httpd_pxe_host and change this var to the http_store

inventory.yml.sample
halleelyasaf: I will delete that

roles/insert_dns_records/tasks/configure_firewall.yml
halleelyasaf: 81 I will delete because we don’t use that we are using the http_store

inventory.yml.sample
```diff
@@ -176,7 +177,14 @@ all:
 
         http_store:
           ansible_host: 10.40.0.100
+        
+        tftp_host
+          ansible_host: 10.40.0.100
+          tftp_directory: /var/lib/tftpboot/
```
arjuhe, line 183: tftp_host shouldn't be here as well. At least not for the default sample inventory.

playbooks/deploy_ipmitool.yml
```diff
@@ -0,0 +1,7 @@
+- name: Deploy IPMI tools
+  hosts: bastion
+  vars:
+    SETUP_IPMI_TOOLS: "{{ setup_pxe_service | default(True) }}"
```
arjuhe, line 4: This will cause SETUP_IPMI_TOOLS to be true if it is not defined. Is this expected?

roles/setup_tftp/handlers/main.yml
```diff
@@ -0,0 +1,14 @@
...
+    name: tftp.socket
+    state: restarted
+  become: true
+
+- name: Start TFTP
+  ansible.builtin.service:
+    name: tftp.socket
+    state: started
+    enabled: true
+  become: true
```
arjuhe, line 13: Please move the `become` to the top of the block. This makes it much easier to determine which block have elevated privileges. 
```
- name: Restart TFTP
   become: true
   ansible.builtin.service
```

roles/setup_tftp/tasks/main.yml
arjuhe: +1

inventory.yml.sample
natifridman: I think we can keep it here for now, since there are some things besides this that we should move from the sample inventory.
This PR is big enough and this can be done as part of the inventory cleanup.

halleelyasaf: > I think we can keep it here for now, since there are some things besides this that we should move from the sample inventory. This PR is big enough and this can be done as part of the inventory cleanup.

Agree

mkochanowski: Initially, the sample inventory file was a minimal working example -- something you can pick up and extend if needed, using the inventory docs. It doesn't mean that we can't change the purpose of the file, of course. 

> This PR is big enough and this can be done as part of the inventory cleanup.

I'd argue that it's better to add the required changes directly in this PR, we never know if priorities will remain the same, and so the cleanup could get delayed. Also, removing the host would actually make the PR smaller!

playbooks/deploy_ipmitool.yml
halleelyasaf: In the inventory is set to False

inventory.yml.sample
natifridman: @mkochanowski Where would you put the documentation about this feature? it still needs to be documented somewhere (with an example)

playbooks/deploy_ipmitool.yml
natifridman: @halleelyasaf What if it doesn't exist in the inventory? you still need to set it to false in the playbook itself

halleelyasaf: OK

deploy_cluster.yml
halleelyasaf: The when is inside the playbook
`when: MOUNT_DISCOVERY_ISO == True`

inventory.yml.sample
mkochanowski: @natifridman I think the Inventory Docs would be a good place (https://github.com/redhat-partner-solutions/crucible/blob/main/docs/inventory.md). We could add a section about configuring alternative boot options (PXE), and maybe link to it from the main README. What do you think, @nocturnalastro?

playbooks/deploy_ipmitool.yml
```diff
@@ -0,0 +1,7 @@
+- name: Deploy IPMI tools
+  hosts: bastion
```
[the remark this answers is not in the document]
nocturnalastro: Don't forget to delete this playbook and role as they are no longer installing the things for use and we can not use become.

inventory.yml.sample
nocturnalastro: So I did come check to see if `tftp_host` is required. Turns out it not because ansible will skip playbooks if there is no host found for them. 

So we could remove from the base inventory. 

It defiantly should be documented in `inventory.md`

We are missing some inventory validations for checks if `pxe` is a vendor we should check that `tftp_host` is defined and the required values are in the `dns_host`

nocturnalastro: If we wish to have sample inventory with `tftp_host` we could always have a inventory sample for pxe specificity. Perhaps its time to make a samples dir.

nocturnalastro (changes requested):
We still have some sudo calls on the bastion.
Also missing inventory validation for catching missing values in the pxe use case.

playbooks/deploy_ipmitool.yml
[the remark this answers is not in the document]
arjuhe: We need to document what software is required on the bastion.

inventory.yml.sample
nocturnalastro: Add validation  in the inventory validation playbook that if `tftp_host` is defined in the inventory. That it should be equal to the `dns_host`. This should also be documented

inventory.yml.sample
```diff
@@ -45,6 +45,7 @@ all:
     # You will need to ensure alternatives are available for anything that will not be automatically set up
     setup_ntp_service: false
     setup_dns_service: false
+    setup_pxe_service: false
```
nocturnalastro, line 48: Validate that if any vendors have `pxe` that the required fields e.g. `tftp_host` is present in the inventory validation playbooks.

inventory.yml.sample
halleelyasaf: Done

playbooks/mount_discovery_iso_for_pxe.yml
```diff
@@ -0,0 +1,6 @@
+- hosts: bastion
```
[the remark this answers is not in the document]
halleelyasaf: Done

roles/mount_discovery_iso_for_pxe/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+        - "/mnt/iso"
+
+    - name: Copy system file to tftp directory
+      copy:
+        src: "{{ item }}"
+        dest: "{{ tftp_root }}"
+        mode: '0644'
+      with_items: "{{ system_files }}"
+
+  become: true
```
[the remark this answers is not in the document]
halleelyasaf: Done

playbooks/deploy_dns.yml
```diff
@@ -3,6 +3,7 @@
   gather_facts: "{{ (setup_dns_service | default(True)) | bool }}"
   vars:
     SETUP_DNS_SERVICE: "{{ setup_dns_service | default(True) }}"
+    TFTP_ROOT: "{{ hostvars['tftp_host']['tftp_directory'] | default('/var/lib/tftpboot/') }}"
```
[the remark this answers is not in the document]
halleelyasaf: Done

playbooks/deploy_ipmitool.yml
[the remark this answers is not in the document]
nocturnalastro: @arjuhe its here https://github.com/redhat-partner-solutions/crucible/pull/93/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R58

roles/validate_inventory/tasks/pxe.yml
```diff
@@ -0,0 +1,33 @@
...
+  changed_when: False
+
+- name: Record failures
+  fail:
+    msg: "ipmitool must be installed for pxe installatio"
+  when: ipmitool_check.rc != 0
+
+- name: "Check if pyghmi installed"
+  ansible.builtin.shell:
+    cmd: "pip3 show pyghmi"
```
natifridman: @nocturnalastro Ansible modules are idempotent (usually), which means that if you tell Ansible to configure something that is already configured it won't do anything. 
Why use the shell module instead of the pip module for making sure the required packages are installed? it is always better to use modules instead of shell commands when possible

natifridman: btw there is a [module](https://docs.ansible.com/ansible/latest/collections/community/general/python_requirements_info_module.html#return-values) that does this check.
If the `not_found` value isn't empty - the package isn't installed

playbooks/deploy_ipmitool.yml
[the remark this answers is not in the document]
halleelyasaf: I will delete this play move to validate_pxe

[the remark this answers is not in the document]
halleelyasaf: DONE

roles/validate_inventory/tasks/pxe.yml
halleelyasaf: We will open new PR for changing all the ensure to 3 way 
1. Check if the package install
2. Install the package (if have sudo access)
3. tell to install (doesn’t have sudo access)

natifridman (approved):
LGTM

README.md
```diff
@@ -55,6 +55,9 @@ Requires the following to be installed on the deployment host:
 - [skopeo](https://github.com/containers/skopeo)
 - [podman](https://github.com/containers/podman/)
 - [kubectl + oc](https://docs.openshift.com/container-platform/4.9/cli_reference/openshift_cli/getting-started-cli.html)
+- [pyghmi](https://pypi.org/project/pyghmi/) #For PXE deployment
+- [ipmitool](https://github.com/ipmitool/ipmitool) #For PXE deployment
```
nocturnalastro, line 59: Can you add `python3-pyghmi` and `ipmitool` to line 67 the `dnf install`

roles/validate_inventory/tasks/validate_pxe.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: Assert 'dns host' and 'tftp host' is have IPs and that is the same
+  assert:
+    that:
+      - hostvars['dns_host']['ansible_host'] is defined
+      - hostvars['tftp_host']['ansible_host'] is defined
+      - hostvars['dns_host']['ansible_host'] == hostvars['tftp_host']['ansible_host']
+    quiet: true
+  when: hostvars['dns_host']['use_pxe'] | default(false)
+
+- name: "Check if ipmitool installed"
```
nocturnalastro, line 11: I don't think this is the correct place for these checks. 
They're not checking inventory values. 
Also validate_inventory runs on `localhost` not on the `bastion` which SHOULD be the same but there not guaranteed. 
I think we need a validate_bastion playbook or something like that.

nocturnalastro: Also this needs a check for if any of the nodes have pxe as we only need them installed if pxe is a vendor.

README.md
halleelyasaf: DONE

roles/validate_inventory/tasks/validate_pxe.yml
halleelyasaf: Can I add delegate to?

halleelyasaf: > Also this needs a when condition for if any of the nodes have pxe as we only need them installed if pxe is a vendor.

We have this check in the main.yml

nocturnalastro: I don't think `setup_pxe_service` Is the correct value here. We could use an existing pxe server, you need to check the vendor.

halleelyasaf: But if we will add the check that when you have `vendor=pxe` that have to use also the `setup_pxe_service =true` it will be the same not?

halleelyasaf: I added the delegation to bastion

inventory.yml.sample
halleelyasaf: If you want to use your own PXE server it will fail

nocturnalastro (approved):
:fireworks:

playbooks/README.md
```diff
@@ -18,6 +19,7 @@ and are nominally put in the order of usage with some exceptions. This may be us
 | `create_cluster.yml`                   | Creates cluster definition in assisted installer              | -                     |
 | `generate_discovery_iso.yml`           | Generate descovery iso                                        | `cluster_id`          |
 | `boot_iso.yml`                         | Reboots nodes to discovery iso                                | -                     |
+| `mount_discovery_iso_for_pxe.yml`      | crate the files for the PXE                                   | -                     |
```
mkochanowski, line 22: Consider changing to something like: 
"Creates files for PXE, mounts discovery iso and configures grub"

mkochanowski (changes requested):
Looks good! Let me propose a few finishing touches to the documentation.

docs/inventory.md
```diff
@@ -218,6 +218,7 @@ Use the following vars to control setup of prerequisites:
 - `setup_dns_service`
 - `create_vms`
 - `setup_sushy_tools`
+- `deploy_tftp.yml`
```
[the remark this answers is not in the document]
halleelyasaf: yes you right

playbooks/README.md
nocturnalastro: I've just noticed something can you move this line above `boot_iso` to preserve the order of execution

README.md
```diff
@@ -164,6 +167,7 @@ Crucible can automatically set up the services required to deploy and run a clus
 - HTTP Store - Used to serve the Assisted Installer discovery ISO and allow it to be used as Virtual Media for nodes to boot from.
 - DNS - Optionally set up DNS records for the required cluster endpoints, and nodes. If not automatically set up then the existing configuration will be validated.
 - Assisted Installer - A pod running the Assisted Installer service, database store and UI. It will be configured for the target environment and is used by the cluster deployment playbook to coordinate the cluster deployment.
+- TFTP Host - A server that stores all the file mounted from the discovery image.
```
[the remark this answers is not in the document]
halleelyasaf: DONE

docs/inventory.md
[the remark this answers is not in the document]
halleelyasaf: DONE

playbooks/README.md
halleelyasaf: I agree that is not a good name but this play does:
1. Mount the iso.
2. Upload the files to the http-store.
3. Make the grub file.
4. Download the files to the TFTP server
Maybe we will split that to more plays

mkochanowski: It's true that the name of the playbook and role is less than perfect, but I don't have a good suggestion on how to fix it.

I think the least what we could do is to provide as much clarification in the description as possible. Based on the flow you've described, what about changing this entry in the table to: 
```
`mount_discovery_iso_for_pxe.yml` | Mounts discovery iso, generates grub config and syncs files to the TFTP server
```

mkochanowski: Please let me know if you think the proposed description is an improvement. This request for changes has the lowest priority and I'm happy to drop it if you disagree with the direction.

halleelyasaf: DONE

arjuhe (approved):
+1 Congrats @halleelyasaf looks good to me. (:

arjuhe:
@halleelyasaf This looks great. All that remains is to squash all the commits into 1 and resolve the conflict in `readme.md`.

Issue #95, opened 2022-03-03
"fetched" directory is not created before it is referenced


abays:
### Bug description

When running `deploy_cluster.yml` one can see that the `create_cluster.yml` playbook is imported and executed before the `install_cluster.yml` playbook:

https://github.com/redhat-partner-solutions/crucible/blob/f32b46317e5bb25fffe16c93fdc0f87a72d423bf/deploy_cluster.yml#L5
https://github.com/redhat-partner-solutions/crucible/blob/f32b46317e5bb25fffe16c93fdc0f87a72d423bf/deploy_cluster.yml#L13

However, the `fetched` directory used by `create_cluster` is not created until later in `install_cluster` ([here](https://github.com/redhat-partner-solutions/crucible/blob/f32b46317e5bb25fffe16c93fdc0f87a72d423bf/roles/install_cluster/tasks/main.yml#L16-L22)), thus resulting in an error like this during execution:

```
TASK [create_cluster : Save cluster_id] ****************************************
Thursday 03 March 2022  10:48:49 +0000 (0:00:00.019)       0:00:05.621 ******** 
fatal: [bastion]: FAILED! => changed=false 
  ...
  msg: Destination directory /home/redhat/crucible//fetched does not exist
```

The `fetched` directory should be created before it is used, perhaps just before this failing tash in `create_cluster`.

### OpenShift version

4.8

### Assisted Installer version

v1.0.24.2

### Required statements

- [X] I have removed all sensitive details from the attached logs and inventory files.
- [X] I acknowledge that Red Hat does not provide commercial support for the content of this repository.
- [X] I acknowledge that any assistance is offered purely on a best-effort basis, as resource permits.

PR #99, opened 2022-03-07
Ensure 'fetched' dir exists before use

abays:
Potential fix for https://github.com/redhat-partner-solutions/crucible/issues/95

### DIFF
--- a/roles/generate_ssh_key_pair/tasks/main.yml
+++ b/roles/generate_ssh_key_pair/tasks/main.yml
@@ -4,6 +4,14 @@
     path: "{{ key_pair_dir }}"
     state: absent
 
+- name: "Make sure {{ fetched_dest }} exists"
+  file:
+    path: "{{ fetched_dest }}"
+    state: directory
+    recurse: yes
+  delegate_to: localhost
+  become: no
+
 - name: Make key dir
   file:
     path: "{{ key_pair_dir }}"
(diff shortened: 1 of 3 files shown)


roles/generate_ssh_key_pair/tasks/main.yml
```diff
@@ -4,6 +4,12 @@
     path: "{{ key_pair_dir }}"
     state: absent
 
+- name: "Make sure {{ fetched_dest }} exists"
+  file:
+    path: "{{ fetched_dest }}"
+    state: directory
+    recurse: yes
```
nocturnalastro, line 11: Hmm we've been a bit fast an lose with `bastion` and `localhost` throughout the playbooks. Can we still delegate this to `localhost`. While we strongly suggest they should be the same there is no guarantee they are. I'll take a note to check the playbooks over to check that the distinction is correct and write up a contributor note to explain the difference.

abays: Ack, will add the delegate

snecklifter (approved):
LGTM, I have also hit that from memory.

roles/generate_ssh_key_pair/tasks/main.yml
arjuhe: @abays thanks for adding that. Coincidentally I brought up the bastion/localhost thing in a separate PR earlier today. 😄

arjuhe (changes requested):
Just add the delegate and we are good to go.

nocturnalastro (approved):

PR #101, opened 2022-03-14
Remove alias for kvm node hostvars

nocturnalastro:
This causes all values in hostvars to be evaluated, this could cause some values to be evaluated
before the value for them is populated e.g. the file it reads does not exist at that point in
the execution


arjuhe (approved):
bummer about aliases.

PR #102, opened 2022-03-14
Provide non-root user the ability to create VMs

hhamid93:
This causes non-root user with sudo privileges to have the ability to create vms. Please review and let me know any further changes are required.


arjuhe (changes requested):
Looks good to me but please squash all the commits.

nocturnalastro (approved):

arjuhe (approved):
Thanks for the squash.
LGTM!

PR #104, opened 2022-03-16
Fix hardcode http port

halleelyasaf:
Fix the hard-coded port 80 make it a variable


roles/validate_http_store/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
 
 - name: Retrieve file from http_store
   uri:
-    url: "{{ http_host }}/{{ test_file_name }}"
+    url: "{{ http_host }}:{{ http_port }}/{{ test_file_name }}"
```
nocturnalastro, line 15: Move http_port into http_host so that it doesn't get added twice

roles/setup_http_store/defaults/main.yml
```diff
@@ -2,6 +2,7 @@ http_store_container_name: http_store
 http_store_pod_name: http_store_pod
 http_dir: /opt/http_store
 http_data_dir: "{{ http_dir }}/data"
+http_port: "{{ hostvars['http_store']['http_store_port'] }}"
```
nocturnalastro, line 5: default to 80 here that way we don't make a breaking change to people's inventories net time they pull

halleelyasaf: I’ll do that.

inventory.yml.sample
```diff
@@ -179,6 +179,7 @@ all:
 
         http_store:
           ansible_host: 10.40.0.100
+          http_store_port: 80
```
arjuhe, line 182: `http_store_port` should be in defaults. Users can add it here as an override but this shouldn't be the source of the var.

roles/setup_http_store/defaults/main.yml
arjuhe: +1

roles/setup_http_store/tasks/main.yml
```diff
@@ -9,7 +9,7 @@
 
     - name: Open http port, zone internal and public, for firewalld
       firewalld:
-        port: "80/tcp"
+        port: "{{ http_port }}/tcp"
```
arjuhe, line 12: Did you mean to use `http_store_port` here?  I don't see where http_port defined.

arjuhe (changes requested):
Just a few clarifications and changes.

inventory.yml.sample
arjuhe: As long as this change isn't going to break anything we can allow it. Afterwards take a pass and determine if any efficiencies can be added here.

roles/validate_http_store/tasks/main.yml
halleelyasaf: DONE

roles/setup_http_store/tasks/main.yml
halleelyasaf: Its in the default of the role

nocturnalastro (approved):

natifridman (approved):
Looking good!

roles/validate_http_store/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 http_store_dir : "{{ iso_download_dest_path | default('/opt/http_store/data') }}"
-http_host: "{{ discovery_iso_server | default('http://' + hostvars['http_store']['ansible_host']) }}"
+http_port: 80
+http_host: "{{ discovery_iso_server | default('http://' + hostvars['http_store']['ansible_host']) }}:{{ http_port }}"
```
nocturnalastro, line 3: You need to put the http_port into the default otherwise it will get added twice as it would have to be in `discovery_iso_server`
```suggestion
http_host: "{{ discovery_iso_server | default('http://' + hostvars['http_store']['ansible_host'] + ':' + http_port') }}"
```

PR #106, opened 2022-03-18
Removing hard-coded sushy-tools path

thekad:
You can set the `sushy_dir` variable, but the service file still pointed
to a static `/opt/sushy-tools` path.


nocturnalastro (approved):

mkochanowski (approved):

arjuhe (commented):
Good catch. +1

PR #107, opened 2022-03-21
Add listen_local flag to dnsmaqs

nocturnalastro:
When set to false, 127.0.0.1 will not be added to listen-address in dnsmasq

### DIFF
--- a/roles/insert_dns_records/defaults/main.yml
+++ b/roles/insert_dns_records/defaults/main.yml
@@ -21,3 +21,4 @@ use_dhcp: false
 dhcp_lease_time: 24h
 
 listen_address: "{{ ansible_default_ipv4.address }}"
+listen_local: true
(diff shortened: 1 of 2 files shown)


roles/insert_dns_records/defaults/main.yml
```diff
@@ -21,3 +21,4 @@ use_dhcp: false
 dhcp_lease_time: 24h
 
 listen_address: "{{ ansible_default_ipv4.address }}"
+listen_local: true
```
arjuhe, line 24: Add a new line to end of file.

roles/insert_dns_records/templates/openshift-cluster.conf.j2
```diff
@@ -2,7 +2,7 @@ domain={{ domain }}
 {% if write_dnsmasq_config %}
 domain-needed
 bogus-priv
-listen-address=127.0.0.1,{{ listen_address }}
+listen-address={% if listen_local %}}127.0.0.1,{% endif %}{{ listen_address }}
```
arjuhe, line 5: Think there is an extra curly here`{% if listen_local %}` ?

PR #111, opened 2022-03-22
Fixes issues with port parsing

snecklifter:
It seems there is an issue with later podman and its being
interpreted incorrectly. So we add quotes to ensure the colon
is part of the string.


snecklifter:
Upstream ref: https://github.com/containers/ansible-podman-collections/issues/394

nocturnalastro (approved):

mkochanowski (approved):

PR #112, opened 2022-03-23
Deploy virtual network before dns

p3ck:
If using the network bridge specified in the virtual environment the dns
role will reference it before it's been created which will cause dnsmasq
to fail to run.  This will break all dns lookups until fixed.

This PR deploys the virtual resources before the dns role.

### DIFF
--- a/deploy_prerequisites.yml
+++ b/deploy_prerequisites.yml
@@ -12,10 +12,10 @@
 
 - import_playbook: playbooks/deploy_registry.yml
 
+- import_playbook: playbooks/create_vms.yml
+
 - import_playbook: playbooks/deploy_dns.yml
 
 - import_playbook: playbooks/deploy_assisted_installer_onprem.yml
-
-- import_playbook: playbooks/create_vms.yml
 
 - import_playbook: playbooks/deploy_sushy_tools.yml


arjuhe (approved):
LGTM

nocturnalastro:
Thanks for opening the PR,

I'm pretty sure this will run into an issue with libvirt also starting up its own dnsmasq instance which will block the dnsmasq created by the dns role. At least that's what I found to happen when another user ran create_vms before the deploy_prerequisites playbook. This also has the added benefit of allowing users test network configurations more easily by just running that playbook. 

I'm about to push a PR which factors out the setting up of the host network config from the `create_vms` playbook and moves just that part before the dns role. Sorry I forgot to create the PR on Monday when I made the changes. I could have saved you some time but I wanted to do some more testing as I overhauled the playbooks for vms in the process. 

For reference I've now open a draft here #113

nocturnalastro:
Opp Sorry I didn't mean to close it I clicked the wrong button. I'll discuss with the rest of the team what we'll do in terms of merging yours vs waiting for mine to be ready.

nocturnalastro (approved):

PR #114, opened 2022-03-24
Install manual binaries in /usr/local/bin

thekad:
One of the requirements says to install the `openshift-clients` package,
which provides `/usr/bin/{kubectl,oc}`. Then when populating the
registry we download the binaries according to the version to be
installed, then overwrite them on top of the ones in `/usr/bin`. It's
probably better to keep the integrity of the RPM DB and install them to
`/usr/local/bin` instead.

### DIFF
--- a/roles/populate_mirror_registry/tasks/populate_registry.yml
+++ b/roles/populate_mirror_registry/tasks/populate_registry.yml
@@ -22,7 +22,7 @@
 
     - name: Mirror remote ocpmetal image registry to local
       command: >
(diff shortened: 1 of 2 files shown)


nocturnalastro:
I agree. Another way we could do it just check if they exist already and skip the steps of getting them.

thekad:
Well, if you download a new version of OCP then you'd preferably want the clients matching the version, so downloading seems fine to me.

nocturnalastro (approved):

PR #116, opened 2022-03-25
Configure registry cert on assisted installer host if using local reg…

nocturnalastro:
…istry


roles/setup_assisted_installer/tasks/main.yml
```diff
@@ -1,12 +1,43 @@
...
+            mode: 0755
+
+        - name: Copy pull_secrets file.
+          copy:
+            content: "{{ pull_secret }}"
+            dest: "{{ config_file_path }}/containers/auth.json"
+            mode: 0644
+            remote_src: yes
+  
+  when: use_local_mirror_registry | bool
```
demvass, line 36: Podman installation is now bounded with the local_mirror_registry flag. Is that what we want here?

roles/setup_mirror_registry/tasks/set_mirror_cert.yml
```diff
@@ -5,5 +5,6 @@
 - name: Populate mirror_certificate in bastion
   set_fact:
     mirror_certificate: "{{ mirror_certificate }}"
-  delegate_to: bastion
+  delegate_to: "{{ item }}"
   delegate_facts: true
+  loop: "{{ cert_targets }}"
```
natifridman, line 10: There is a missing newline at the end of the file

roles/setup_assisted_installer/defaults/main.yml
```diff
@@ -85,3 +85,7 @@ assisted_installer_hardware_validation:
       ram_mib: 32768
       disk_size_gb: 120
       installation_disk_speed_threshold_ms: 10
+
+use_local_mirror_registry: "{{ setup_registry_service | default(False) }}"
+
+config_file_path: /tmp/wip/config
```
natifridman, line 91: Do we want the default to be in tmp folder? Should it be in another folder? Something like /opt

roles/setup_assisted_installer/tasks/main.yml
nocturnalastro: Good spot that is at the wrong indent :)

roles/setup_assisted_installer/defaults/main.yml
nocturnalastro: I used `/tmp/wip/config` to match the setup registry code. It should only be required for pulling the containers then it wont be used any more. If we want to clean this up I would prefer to do both in a different PR

arjuhe: I recommend we stop using /tmp and instead create a crucible/tmp folder or .crucible folder in the home directory.

arjuhe (approved):

roles/setup_assisted_installer/defaults/main.yml
nocturnalastro: This is truly a tmp directory though. XDG_RUNTIME_DIR is pointed at it during the pulling of the images as it contains the containers auth file.

PR #118, opened 2022-03-30
Update logic in sushy-tools to use stat

p3ck:
Instead of relying on redhat release version which doesn't work with
centos stream 8, check on the actual file existing instead.

### DIFF
--- a/roles/setup_sushy_tools/tasks/main.yml
+++ b/roles/setup_sushy_tools/tasks/main.yml
@@ -70,6 +70,22 @@
             privatekey_path: "{{ sushy_cert_dir }}/sushy_tools.key"
             cert_common_name: "{{ sushy_fqdn }}"
 
+    - name: Set X86_64 OVMF code path
+      set_fact:
+        sushy_x86_64_ovmf_code_path: "/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd"
+
+    - name: Check x86_64 OVMF code path
+      stat:
+        path: "/usr/share/OVMF/OVMF_CODE.secboot.fd"
+      register: OVMF_X86_64_CODE_STAT
+
+    - name: Set x86_64 OVMF code path (legacy)
+      set_fact:
+        sushy_x86_64_ovmf_code_path: "/usr/share/OVMF/OVMF_CODE.secboot.fd"
+      when:
+        - OVMF_X86_64_CODE_STAT.stat.exists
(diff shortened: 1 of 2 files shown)


nocturnalastro (approved):

roles/setup_sushy_tools/vars/main.yml
```diff
@@ -0,0 +1 @@
+sushy_x86_64_ovmf_code_path: "/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd"
```
natifridman, line 1: I think you can put it in the defaults directory

p3ck: I originally had it in defaults and moved it because it felt a little dirty to me.  Correct me if I'm wrong but defaults feels like it should be, well, a default and not then overwritten by the playbook logic.

Let me know what you decide and if needed I can move it to defaults.

p3ck: to clarify, defaults to me says the user can pass in a different path, but it could still be overwritten by the logic.

thekad: if it's all internal logic, why put it anywhere in defaults or vars? just handle it all in the task itself imo

p3ck: ok - I put all the logic in the tasks

arjuhe: @p3ck, you bring up an interesting topic. The vars file will always overwrite the defaults file in term of precedence. That being said vars files should only be used by users looking for overrides. What do you think?

p3ck: I really don't understand the logic behind ansible sometimes. :-)

I would have expected vars to be private to the role and defaults to be sane defaults that the user should be ok with overriding.

This article seems sane to me.
https://www.bouledef.eu/~tleguern/articles/ansible-private-variables/en/

Any private vars should be prepended with __ to signify that they are not to be overridden.  We have also prepended variables with the role name to keep from clashing with other roles vars.  so something like this would make sense to me:

__<ROLENAME>_<VARNAME>

This is just a suggestion.  I'm not bothered too much.  I hope the updated PR here is satisfactory.

p3ck:
Ping - Need anything else from me?

natifridman (approved):

PR #120, opened 2022-04-05
Add greater controll over listening interfaces for dnsmasq

nocturnalastro:


roles/insert_dns_records/defaults/main.yml
```diff
@@ -21,3 +21,6 @@ use_dhcp: false
 dhcp_lease_time: 24h
```
arjuhe, line 21: dhcp_lease_time should be reduced to avoid lease kruft.

arjuhe (approved):
Just a small change but not a blocker.

roles/insert_dns_records/defaults/main.yml
nocturnalastro: I'll make a new PR for that :)

docs/inventory.md
```diff
@@ -255,6 +255,35 @@ If you need VMs with static UUIDs to allow them to be reused then the UUID for a
 
 By default an SSH key will be generated by the `deploy_cluster.yml` playbook. This can be disabled by adding `generate_ssh_keys = False` to the inventory. It is possible to configure the task generating the SSH key (see the docs for `community.crypto.openssh_keypair`) by setting `openssh_keypair_args` with a dictionary.
 
+### DNS and DHCP
+
+If you to only point to the dns configure by crucible you should also provide `upstream_dns` which points to another dns server which can provide records for non-crucible queries.
+
+You can control the ip addresses which dnsmasq listens to using the `listen_addresses` by default this will include both `120.0.0.1` and ansible's default IPv4 address for the host (`ansible_default_ipv4.address`). You may also configure the interfaces which dnsmasq responds by defining `listening_interfaces` as a list of the interfaces you which to listen to.
```
natifridman, line 262: If I want to define the listening addresses, should I use `listen_addresses` or `listening_interfaces` in the inventory?
Maybe it is worth adding an example for both of them

roles/insert_dns_records/templates/openshift-cluster.conf.j2
```diff
@@ -2,7 +2,13 @@ domain={{ domain }}
 {% if write_dnsmasq_config %}
 domain-needed
 bogus-priv
-listen-address=127.0.0.1,{{ listen_address }}
+listen-address="{{ listen_addresses | join(',') }}"
+{% for  listening_intf in (listening_interfaces | default([])) %}
+interface={{ listening_intf }}
```
natifridman, line 7: Don't we need also to tell dnsmasq which interface to use for the DHCP range in case we use more than one interface?

nocturnalastro: No from what I can see of the config it is done via disabling DHCP on a given interface `no-dhcp-interfac`. Or are you saying having an interface per dhcp range?

natifridman: I meant to have dhcp range per interface

docs/inventory.md
nocturnalastro: Added example

natifridman (approved):

PR #123, opened 2022-04-13
Remove the need to define a tftp_host

nocturnalastro:
Add a default to tftp_host in TFTP_ROOT


halleelyasaf (approved):
LGTM

natifridman:
Is there any way to make it more readable? It took me some time to understand what is going on here

nocturnalastro:
Unfortunately aliasing hostvars in such a way would lead to all values being evaluated which has side effects.

natifridman (approved):
LGTM

arjuhe (approved):

PR #124, opened 2022-04-13
Fix gather_facts ignoring inventory variables

cdonatom:
Small fix for gather_facts instruction. It now follows the conditions to trigger fact gathering.

### DIFF
--- a/playbooks/create_vms.yml
+++ b/playbooks/create_vms.yml
@@ -8,9 +8,13 @@
 
 - name: Provision VMS
   hosts: vm_hosts
-  gather_facts: "{{ (setup_vms | default((kvm_nodes | default([])) | length | int >= 1)) | bool }}"
+  gather_facts: False
   vars:
     SETUP_VMS: "{{ setup_vms | default((kvm_nodes | default([])) | length | int >= 1) }}"
+  pre_tasks:
+    - name: Setup facts
+      setup:
+      when: SETUP_VMS == True
   roles:
     - role: destroy_vms
       when: SETUP_VMS | bool
--- a/playbooks/deploy_ntp.yml
+++ b/playbooks/deploy_ntp.yml
@@ -1,8 +1,12 @@
 - name: Setup NTP
   hosts: ntp_host
-  gather_facts: "{{ (setup_ntp_service | default(True)) | bool }}"
+  gather_facts: False
   vars:
     SETUP_NTP_SERVICE: "{{setup_ntp_service | default(True)}}"
+  pre_tasks:
+    - name: Setup facts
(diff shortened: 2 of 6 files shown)


playbooks/deploy_ntp.yml
```diff
@@ -1,8 +1,12 @@
 - name: Setup NTP
   hosts: ntp_host
-  gather_facts: "{{ (setup_ntp_service | default(True)) | bool }}"
+  gather_facts: False
   vars:
     SETUP_NTP_SERVICE: "{{setup_ntp_service | default(True)}}"
   roles:
     - role: setup_ntp
       when: SETUP_NTP_SERVICE == True
+  pre_tasks:
```
nocturnalastro, line 9: Can you put the pre_tasks before the roles.

arjuhe (approved):

playbooks/deploy_ntp.yml
arjuhe: Oh snap! I can't believe I missed this one here. @nocturnalastro good catch.

nocturnalastro:
Can you also add assisted installer looks like it was missed when adding the original flags.

playbooks/create_vms.yml
```diff
@@ -24,3 +24,7 @@
       when: SETUP_VMS | bool
       tags:
         - setup_vms
+  pre_tasks:
```
arjuhe, line 27: Just link @nocturnalastro said. Please put the pre_tasks before the roles.

arjuhe (changes requested):

playbooks/deploy_ntp.yml
cdonatom: There you go! Thx for the feedback!

playbooks/deploy_registry.yml
```diff
@@ -1,11 +1,15 @@
 ---
 - name: Play to populate image_hashes for relevant images
   hosts: localhost
-  gather_facts: "{{ (setup_registry_service | default(True)) | bool }}"
+  gather_facts: False
   vars:
+    SETUP_REGISTRY_SERVICE: "{{ setup_registry_service | default(True)}}"
     destination_hosts:
       - registry_host
   pre_tasks:
```
arjuhe, line 9: This is not a blocker but in this case `pre_tasks` would get replaced with `tasks` since there is no task section.

arjuhe (approved):
Looking good. Please squash your commits.

nocturnalastro (approved):

nocturnalastro:
@cdonatom Could you squash your commits into one here :)

PR #127, opened 2022-04-18
Fix syntax error in inventory sample

p3ck:
tftp_host is missing the : which invalidates the entire yaml file.


arjuhe (approved):
Good catch. I bumped into this last week.

nocturnalastro (approved):

PR #128, opened 2022-04-18
Naked bridge

p3ck:
Minor update to allow fully isolated virt environments which we use for our CI setup.

With this patch we can specify the following in the hosts: -> vm_host1 section.

```yaml
          network_config:
            interfaces:
              - type: linux-bridge
                name: "{{ cluster_name }}-br"
                addresses:
                  ipv4:
                    - ip: "{{ vm_bridge_ip }}"
                      prefix: "{{ machine_network_cidr | ipaddr('prefix') }}"
                bridge:
                  stp: True
```

### DIFF
--- a/roles/process_nmstate/templates/nmstate.yml.j2
+++ b/roles/process_nmstate/templates/nmstate.yml.j2
@@ -54,11 +54,13 @@ interfaces:
         multicast-snooping: {{ interface.bridge.snooping | default(true) }}
         stp:
           enabled: {{ interface.bridge.stp | default(true) }}
+      {% if interface.bridge.port is defined %}
       port:
         - name: {{ interface.bridge.port.name }}
           stp-hairpin-mode: {{ interface.bridge.port.hairpin | default(false) }}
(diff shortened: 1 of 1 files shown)


nocturnalastro:
An Isolated virt env's are probably not what people are aiming for most of the time. So I feel like there should be a validation check here still. Perhaps with an option to turn it off but I think it should be a concious thing. 

Maybe we could use something like setting the port to `None` or `false` as the trigger for an isolated network.

p3ck:
Not sure if you realized or not, but this doesn't change the default behavior.  You have to pass in a network_config for the vm_host and even then you have to omit the port entry.  So I think if one does that then it's a pretty conscious decision.

You allow for customization by allowing for the network_config (nmstate) to be passed in, but force that customization.  This patch simply allows you more customization.

Thanks

nocturnalastro:
Good point :+1: As you say this is pretty deliberate.

nocturnalastro (approved):

arjuhe (approved):
LGTM

PR #129, opened 2022-04-19
Don't process existing OCP artifacts

thekad:
Crucible will download OpenShift artifacts it needs, but it would be
better if it didn't do it for every run.
This is good for repeatable runs where you are testing the same
version(s) over and over again: downloaded artifacts will be reused as
a sort of cache.


nocturnalastro (approved):
Thanks this has been on my TODO this for a while.

arjuhe (approved):
LGTM

PR #130, opened 2022-04-20
Adding stand alone playbook to register hosts and enable repos.

arjuhe:
This playbook is not intended to be used in OCP deployments but a stand alone playbook used to help prepare hosts.

### DIFF
--- a/playbooks/rhsm_register.yml
+++ b/playbooks/rhsm_register.yml
@@ -1,0 +1,44 @@
+---
+- name: Register host with Red Hat Subscription Manager
+  hosts: bastions,services,vm_hosts
+  gather_facts: true
+  vars_prompt:
+    - name: rhsm_username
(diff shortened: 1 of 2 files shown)


playbooks/rhsm_register.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: rhsm_username
+      prompt: "RHSM Username"
+      private: no
+    - name: rhsm_password
+      prompt: "RHSM Password"
+      private: yes
+      unsafe: yes
+  vars:
+    rhel_release: 8.5
+    rhsm_pool_ids: # Can accept lists
```
nocturnalastro, line 15: Default to empty list?

arjuhe: good idea.

nocturnalastro (approved):

PR #131, opened 2022-04-22
Fixing linting issues

nocturnalastro:
Issues are listed here https://github.com/nocturnalastro/crucible/runs/6126261622?check_suite_focus=true

### DIFF
--- a/roles/create_vms/tasks/provision_vms.yml
+++ b/roles/create_vms/tasks/provision_vms.yml
@@ -25,5 +25,7 @@
       loop: "{{ kvm_nodes }}"
 
     - name: Run vm creation_scripts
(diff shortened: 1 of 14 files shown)


roles/create_vms/tasks/provision_vms.yml
```diff
@@ -25,5 +25,7 @@
       loop: "{{ kvm_nodes }}"
 
     - name: Run vm creation_scripts
-      shell: "/bin/bash {{ vm_create_scripts_dir }}/{{ item.name }}_setup_vm.sh"
+      shell:  # noqa: no-changed-when
+        cmd: "/bin/bash {{ vm_create_scripts_dir }}/{{ item.name }}_setup_vm.sh"
```
arjuhe, line 29: I think this will always produce a _changed_ due to the fact it needs to run the shell command.

arjuhe (approved):
+1

Issue #148, opened 2022-06-17
Proxy settings not populated into the AI Pod config


jtudelag:
### Bug description

When the host where the AI Pod is running has access to internet via proxy, that is not taken into account into the AI Pod config, that requires internet access to pull the OCP release container payload (among other things probably):

`time="2022-06-16T14:12:50Z" level=warning msg="Failed to generate steps for command *hostcommands.imageAvailabilityCmd" func="github.com/openshift/assisted-service/internal/host/hostcommands.(*InstructionManager).GetNextSteps" file="/go/src/github.com/openshift/origin/internal/host/hostcommands/instruction_manager.go:158" error="command 'oc adm release info --image-for=machine-config-operator --insecure=false quay.io/openshift-release-dev/ocp-release@sha256:bf48faa639523b73131ec7c91637d5c94d33a4afe09ac8bdad672862f5e86ccb --registry-config=/tmp/registry-config534443934' exited with non-zero exit code 1: \nerror: unable to read image quay.io/openshift-release-dev/ocp-release@sha256:bf48faa639523b73131ec7c91637d5c94d33a4afe09ac8bdad672862f5e86ccb: Get \"https://quay.io/v2/\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)\n" go-id=339830 host_id=1799628f-6d23-340a-79ed-a3bf4822d4aa pkg=instructions request_id=f51fca9c-4cee-4ba5-bcf8-23d17eb60074`

This should be fixed in 

- In the v1 api branch proxy settings should be populated here: https://github.com/redhat-partner-solutions/crucible/blob/crucible-ai-v1-api-final/roles/setup_assisted_installer/templates/onprem-environment.j2
- In the v2 api branch, they should be populated here: https://github.com/redhat-partner-solutions/crucible/blob/main/roles/setup_assisted_installer/templates/configmap.yml.j2

 

### OpenShift version

4.8

### Assisted Installer version

v1.0.24.2

### Relevant log output

```shell
[...]

PR #151, opened 2022-06-21
Fix proxy settings

jtudelag:
Set and populate proxy seettings in:

- Inventory
- AI config
- Day2 clusters config

Also remove unused proxy settings from the patch_cluster role.

Addresses https://github.com/redhat-partner-solutions/crucible/issues/148

### DIFF
--- a/inventory.yml.sample
+++ b/inventory.yml.sample
@@ -37,9 +37,17 @@ all:
     # network_type: OVNKubernetes
     # network_type: OpenShiftSDN
 
+
+
     ######################################
     # Prerequisite Service Configuration #
     ######################################
+
+    # Proxy settings. These settings apply to: Assisted Installer, Day1 clusters and Day2 clusters.
+    # This assumes the host where the AI runs and the OpenShift cluster share the same proxy settings.
+    # http_proxy: ""
+    # https_proxy: ""
+    # no_proxy: ""
 
     # Flags to enable/disable prerequisite service setup
     # You will need to ensure alternatives are available for anything that will not be automatically set up
--- a/roles/create_day2_cluster/defaults/main.yml
+++ b/roles/create_day2_cluster/defaults/main.yml
@@ -12,9 +12,9 @@ ASSISTED_INSTALLER_HOST: "{{ hostvars['assisted_installer']['host'] | default(an
 ASSISTED_INSTALLER_PORT: "{{ hostvars['assisted_installer']['port'] | default(8090) }}"
 ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ ASSISTED_INSTALLER_HOST }}:{{ ASSISTED_INSTALLER_PORT }}/api/assisted-install/v2"
 
-HTTP_PROXY: ""
-HTTPS_PROXY: ""
-NO_PROXY: ""
+http_proxy: ""
+https_proxy: ""
+no_proxy: ""
 
 # HTTP Basic Authentication
 HTTP_AUTH_USERNAME: "none"
--- a/roles/setup_assisted_installer/defaults/main.yml
+++ b/roles/setup_assisted_installer/defaults/main.yml
@@ -42,3 +42,8 @@ use_local_mirror_registry: "{{ setup_registry_service | default(False) }}"
 
 config_file_path: /tmp/wip/config
 port: 8090
+
+# Proxy settings for the AI Pod configuration.
+http_proxy: ""
+https_proxy: ""
+no_proxy: ""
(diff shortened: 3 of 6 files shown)


roles/setup_assisted_installer/templates/configmap.yml.j2
```diff
@@ -32,3 +32,10 @@ data:
   INSTALLER_IMAGE: {{ assisted_service_installer_image }}
   CONTROLLER_IMAGE: {{ assisted_service_controller_image }}
   AGENT_DOCKER_IMAGE: {{ assisted_service_installer_agent_image }}
+  {% if http_proxy is defined %}
+  HTTP_PROXY: {{ http_proxy }}
+  {% elif https_proxy is defined %}
+  HTTPS_PROXY: {{ https_proxy }}
+  {% elif no_proxy is defined %}
+  NO_PROXY: {{ no_proxy }}
+  {% endif %}
```
nocturnalastro, line 41: I'm pretty sure you want all 3 of these together so you don't want to use `elif`

roles/create_day2_cluster/defaults/main.yml
```diff
@@ -12,9 +12,9 @@ ASSISTED_INSTALLER_HOST: "{{ hostvars['assisted_installer']['host'] | default(an
 ASSISTED_INSTALLER_PORT: "{{ hostvars['assisted_installer']['port'] | default(8090) }}"
 ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ ASSISTED_INSTALLER_HOST }}:{{ ASSISTED_INSTALLER_PORT }}/api/assisted-install/v2"
 
-HTTP_PROXY: ""
-HTTPS_PROXY: ""
-NO_PROXY: ""
+http_proxy: ""
+https_proxy: ""
+no_prox: ""
```
nocturnalastro, line 17: missing `y`

inventory.yml.sample
```diff
@@ -37,10 +37,18 @@ all:
...
+
     ######################################
     # Prerequisite Service Configuration #
     ######################################
 
+    # Proxy settings. These settings apply to: Assisted Installer, Day1 clusters and Day2 clusters.
+    # This assumes the host where the AI runs and the OpenShift cluster share the same proxy settings.
+    # http_proxy: ""
+    # https_proxy: ""
+    # no_prox: ""
```
nocturnalastro, line 50: missing `y`

roles/setup_assisted_installer/templates/configmap.yml.j2
nocturnalastro, line 41:
```suggestion
  {% if http_proxy is defined %}
  HTTP_PROXY: {{ http_proxy }}
  {% endif %}
  {% if https_proxy is defined %}
  HTTPS_PROXY: {{ https_proxy }}
  {% endif %}
  {% if no_proxy is defined %}
  NO_PROXY: {{ no_proxy }}
  {% endif %}
```

roles/setup_assisted_installer/defaults/main.yml
```diff
@@ -41,4 +41,8 @@ image_service_host: "{{ host | default('ansible_fqdn') }}" # bastion -> image se
 use_local_mirror_registry: "{{ setup_registry_service | default(False) }}"
 
 config_file_path: /tmp/wip/config
-port: 8090
\ No newline at end of file
+port: 8090
+
+#http_proxy: ""
```
nocturnalastro, line 46: could you remove these :)

nocturnalastro (changes requested):
Thanks for this :) 
There's a couple of tweaks I think need to be made

roles/setup_assisted_installer/defaults/main.yml
jtudelag: Yes, I dont disagree...On the other hand added them there in case someone wants to add the proxy settings, to at least know they can set them for the AI...so they have to start grepping or finding proxy related stuff. Not sure if I am explaining myself..

nocturnalastro: Ah that's what you mean with that! 
At minimum I would add a comment to explain that. 
I think a clearer way is to use a "magic" value like an empty string and instead of checking if it is defined check it not the magic value.

jtudelag: I am ok with that. Just make the changes...PTAL.

nocturnalastro (commented):
Can you squash your commits

jtudelag:
Done @nocturnalastro. Thanks for the review!

arjuhe (approved):
Lgtm

PR #154, opened 2022-06-29
Remove day2 patch cluster and two tiny changes

nocturnalastro:


roles/create_cluster/defaults/main.yml
```diff
@@ -16,8 +16,8 @@ ASSISTED_INSTALLER_PORT: "{{ hostvars['assisted_installer']['port'] | default(80
 ASSISTED_INSTALLER_BASE_URL: "{{ secure | ternary('https', 'http') }}://{{ ASSISTED_INSTALLER_HOST }}:{{ ASSISTED_INSTALLER_PORT }}/api/assisted-install/v2"
 URL_ASSISTED_INSTALLER_CLUSTERS: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters"
 
-http_auth_username: "none"
-http_auth_password: "none"
+HTTP_AUTH_USERNAME: "none"
```
nocturnalastro, line 19: In line it is upper case
https://github.com/redhat-partner-solutions/crucible/blob/887f2c3b210c0833197e26bcf48a3e922eda2a65/roles/create_cluster/tasks/main.yml#L48

arjuhe, line 19: This var should be lower case like it's neighbors.

arjuhe (changes requested):
Just a small adjustment.

roles/create_cluster/defaults/main.yml
nocturnalastro: This was made lower case by accident. In the code is upper case see
https://github.com/redhat-partner-solutions/crucible/pull/154#discussion_r909474487
It is upper case throughout all the other roles as well. 

This just fixing an issue if we want to make it lower case overall I think that should be in another PR as it will be much larger.

arjuhe:
accepted.

PR #158, opened 2022-07-06
Allow for users to defined the password for the core user in discover…

nocturnalastro:
…y iso


docs/inventory.md
```diff
@@ -595,3 +595,7 @@ all:
           ansible_host: 192.168.10.17
           bmc_ip: 172.30.10.7
 ```
+# Defining a password for the discovery iso.
+
+If users wish to provide password for the discovery iso they must define `hashed_discovery_password` in the `all` section inventory.
+The value provided in `hashed_discovery_password` can be created by using `mkpasswd --method=SHA-512 MyAwesomePassword`.
```
arjuhe, line 601: This should be accompanied with a warning.

arjuhe (approved):
LGTM

PR #159, opened 2022-07-06
The split should be done on the `@` character

thekad:
The check is still good, as the `:` character tells you there is a SHA
provided, but skopeo returns `sha256:$SHA`, as it is right now it
results in a malformed image spec with a repeated sha256 bit:
`quay.io/.../ocp-release@sha256@sha256:$SHA`

Also unrelated changes: stripped all dangling whitespace.

### DIFF
--- a/roles/get_image_hash/tasks/main.yml
+++ b/roles/get_image_hash/tasks/main.yml
@@ -1,16 +1,16 @@
 - name: check skopeo is installed
   shell: /usr/bin/skopeo --version
 
-- name: Filter images 
+- name: Filter images
   when: get_release_images | bool
-  block: 
+  block:
     - name: Filter os images
       set_fact:
         os_images: "{{ os_images | json_query(os_filter) }}"
       vars:
         os_filter: "[?(openshift_version == '{{ openshift_version }}')]"
       when: (not get_all_release_versions) | bool
-    
+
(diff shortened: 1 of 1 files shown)


roles/get_image_hash/tasks/main.yml
```diff
@@ -68,24 +68,24 @@
   when: get_release_images | bool
   block:
     - name: Update released items
-      vars: 
+      vars:
         updated_release_images: []
       set_fact:
-        updated_release_images: "{{ updated_release_images + [item | combine({'url': item.url.split(':')[0] + '@' + image_hashes['release_' + item.version + '_' + item.cpu_architecture]  })] }}"
-      when: 
+        updated_release_images: "{{ updated_release_images + [item | combine({'url': item.url.split('@')[0] + '@' + image_hashes['release_' + item.version + '_' + item.cpu_architecture]  })] }}"
```
nocturnalastro, line 74: Ah I this was to replace the tag with the hash. 
Looks like we need a check to make sure the url doesn't already contain the hashed version.

thekad: gotcha, I'll close this PR then

thekad:
will open another PR with the right check for fully qualified image url

PR #160, opened 2022-07-07
Taking into account fully qualified image URLs

thekad:
You can specify the image as `$REGISTRY/$NS/$IMAGE:$TAG` but you can
also specify an image as `$REGISTRY/$NS/$IMAGE@$SHA_ALGO:$SHA_STR` e.g.
how they are set in `release.txt`: "Pull From: $IMAGE"

If the image is fully qualified then there will be an `@` character in
the URL, if that happens then there is no need to re-inspect the image
to get the signature.

### DIFF
--- a/roles/get_image_hash/tasks/main.yml
+++ b/roles/get_image_hash/tasks/main.yml
@@ -67,11 +67,17 @@
 - name: Update release_images with hashes
   when: get_release_images | bool
   block:
-    - name: Update released items (image:tag format)
-      vars:
-        updated_release_images: []
+    - name: Prepopulate list with fully qualified images
       set_fact:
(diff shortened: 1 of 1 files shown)


roles/get_image_hash/tasks/main.yml
```diff
@@ -67,25 +67,26 @@
 - name: Update release_images with hashes
   when: get_release_images | bool
   block:
-    - name: Update released items
-      vars: 
+    - name: Update released items (image:tag format)
+      vars:
         updated_release_images: []
```
nocturnalastro, line 72: I think we need to pre-populate `updated_release_images` with any fully qualified images, otherwise they will be dropped by the loop.

arjuhe: +1

thekad: pushed another commit, lmk if that is what you had in mind

nocturnalastro (approved):

thekad:
@arjuhe do you think there are more changes needed here?

PR #161, opened 2022-07-11
Don't process bmc_dns_records if not defined

p3ck:
Pretty self explanatory, If bmc_dns_records isn't defined then don't try and process them.


nocturnalastro (approved):

arjuhe (approved):
+1

PR #163, opened 2022-07-27
Refresh versions and RHCOS images

novacain1:
This PR:

(1) refreshes RHCOS base images used for both the 4.9 and 4.10 releases to a more recent version hosted on the OpenShift mirror site (mirror.openshift.com).  These represent 6 months of RHEL fixes and should be used as a base vs their older counterparts, as known problems have been observed in lab deployments with HA topologies.
(2) refreshes supported OpenShift versions to releases that contain rebases for `ice` and `iavf` 
drivers.  Customers and partners should also try to stay current with enhancements and security fixes present in these newer versions.
(3) update documentation to reflect support for both OpenShift 4.9 and 4.10, as well as links to public OpenShift documentation for the 4.10 release.

### DIFF
--- a/tests/validate_inventory/templates/all_reachable.yml
+++ b/tests/validate_inventory/templates/all_reachable.yml
@@ -2,7 +2,7 @@ all:
   vars:
     setup_ntp_service: false
     ntp_server: localhost
-    openshift_full_version: 4.6.16
+    openshift_full_version: 4.10.16
 
   children:
     bastions:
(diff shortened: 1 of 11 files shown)


nocturnalastro (approved):

tests/validate_inventory/templates/all_reachable.yml
```diff
@@ -2,7 +2,7 @@ all:
   vars:
     setup_ntp_service: false
     ntp_server: localhost
-    openshift_full_version: 4.6.16
+    openshift_full_version: 4.10.16
```
nocturnalastro, line 5: Changing the version in the tests is proably overkill

arjuhe (approved):
+1

PR #167, opened 2022-07-29
More robust install cluster and small fixes

nocturnalastro:
Make install cluster more robust by allowing cluster the cluster state to be updated before the check.


arjuhe (approved):
LGTM +1

PR #171, opened 2022-08-05
Changing service network and pod

douglasrasilva:
After installing a cluster, I noticed that the pod's network was not changed as configured in the inventory, setting the default network to 10.128.0.0/14.

After adding the two lines below, the following installation correctly configured the pod's network.

We add the lines below:
"cluster_networks": [{"cidr": cluster_network_cidr, "host_prefix": (cluster_network_host_prefix | int)}],
"service_networks": [{"cidr": service_network_cidr}],

Changed file:
"roles/create_cluster/tasks/main.yml"


nocturnalastro:
Hey Douglas sorry for the delay. Thanks for the PR. This functionality has already been added in PR #168 in a different part of the code.

PR #173, opened 2022-08-24
Add list ntp_server_allows

nocturnalastro:
Defaults to [ntp_server_allow] if ntp_server_allow is defined.
This will be required to allow dual stack installations to communicate
with chrony. It will also be useful for more lab settings where multiple
clusters are being set up.


arjuhe (approved):
Lgtm

PR #179, opened 2022-09-28
Fix: no pxe hosts found.

arjuhe:
- Fixed an issue where crucible crashes if setup_pxe_service is true and hosts:vendor is not set to pxe.
- The setup of the pxe service should not be tied to whether or not any host will use pxe.

### DIFF
--- a/roles/display_deployment_plan/templates/plan.j2
+++ b/roles/display_deployment_plan/templates/plan.j2
@@ -24,7 +24,12 @@ An OpenShift cluster is about to be deployed. Please double check the provided i
   {{ row_format_str | format('Setup HTTP Store Service', inventory.setup_http_store_service) }}
   {{ row_format_str | format('Setup Assisted Installer', inventory.setup_assisted_installer) }}
   {{ row_format_str | format('Setup PXE Service', inventory.setup_pxe_service) }}
-
+  {% if inventory.setup_pxe_service | bool and inventory.num_pxe_nodes | int > 0 -%}
+  {{ row_format_str | format('Number of PXE Nodes', inventory.num_pxe_nodes) }}
+  {% elif inventory.setup_pxe_service | bool and inventory.num_pxe_nodes | int == 0 +%}
+  {{ row_format_str | format('Number of PXE Nodes', messages.pxe_hosts_missing) }}
+  {% else %}
+  {% endif +%}
   {{ row_format_str | format('Discovery ISO Name', inventory.discovery_iso_name) }}
   {{ row_format_str | format('Discovery ISO Download Path', inventory.discovery_iso_download_path) }}
 
--- a/roles/validate_inventory/tasks/validate_pxe.yml
+++ b/roles/validate_inventory/tasks/validate_pxe.yml
@@ -8,37 +8,29 @@
     quiet: true
   when: hostvars['dns_host']['use_pxe'] | default(false)
 
-- name: Get all PXE Nodes
-  vars:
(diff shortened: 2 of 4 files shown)


arjuhe:
ERRATA:
TASK [validate_inventory : Record failures] ************************************
fatal: [localhost -> bastion]: FAILED! =>
  msg: |-
    The conditional check 'pxe_node_names | length >= 1' failed. The error was: error while evaluating conditional (pxe_node_names | length >= 1): 'pxe_node_names' is undefined

    The error appears to be in '/<scrubbed>/crucible/roles/validate_inventory/tasks/validate_pxe.yml': line 27, column 5, but may
    be elsewhere in the file depending on the exact syntax problem.

roles/display_deployment_plan/templates/plan.j2
```diff
@@ -24,7 +24,12 @@ An OpenShift cluster is about to be deployed. Please double check the provided i
   {{ row_format_str | format('Setup HTTP Store Service', inventory.setup_http_store_service) }}
   {{ row_format_str | format('Setup Assisted Installer', inventory.setup_assisted_installer) }}
   {{ row_format_str | format('Setup PXE Service', inventory.setup_pxe_service) }}
-
+  {# row_format_str | format('Number of PXE Nodes', inventory.num_pxe_nodes if inventory.num_pxe_nodes not '' else inventory.num_pxe_nodes, "[WARNING] No hosts are set." ) #}
```
nocturnalastro, line 27: Can you remove this line

roles/display_deployment_plan/templates/plan.j2
```diff
@@ -24,7 +24,12 @@ An OpenShift cluster is about to be deployed. Please double check the provided i
...
+  {%- if inventory.num_pxe_nodes | int > 0 %}
+{{ row_format_str | format('Number of PXE Nodes', inventory.num_pxe_nodes) }}
+  {% else %}
+{{ row_format_str | format('Number of PXE Nodes', messages.pxe_hosts_missing) }}
```
nocturnalastro, line 31: Perhaps we want to change the logic here. 
We only need to warn the user there are no pxe nodes and `if setup_pxe_service == true` 
Maybe check for that state and if not just show how many there are.

roles/validate_inventory/tasks/validate_pxe.yml
```diff
@@ -8,37 +8,30 @@
...
-    changed_when: False
-  
-  - name: Record failures
-    fail:
-      msg: "pyghmi must be installed for pxe installation"
-    when: pyghmi_check.rc != 0
-  when: pxe_node_names | length >= 1
+- name: Check for ipmitool and pyghmi
   delegate_to: bastion
+  become: true
```
nocturnalastro, line 13: This can't have `become: true` its executed on the bastion. Also doesn't look like it needs root as its just checking if these packages are installed without trying to install them.

nocturnalastro (changes requested):

roles/display_deployment_plan/templates/plan.j2
arjuhe: please excuse my mess. (:

roles/display_deployment_plan/templates/plan.j2
arjuhe: Warnings only show up when setup_pxe_serice is true and num_pxe_nodes is > than 0.

nocturnalastro (approved):

PR #181, opened 2022-10-04
Add zt vendor

nocturnalastro:


arjuhe (changes requested):
It looks like all files but ../defaults/main.yml need line breaks at the end.

arjuhe (approved):
LGTM

PR #182, opened 2022-10-18
Fix logic issue when skipping task include and overwrite cache file w…

nocturnalastro:
…ith updates


roles/get_image_hash/tasks/get_image_hash.yml
```diff
@@ -1,5 +1,5 @@
 - name: "A hash wasn't provided, inspect the image to figure it out"
-  when: not (item.value.hash | default('')) != ''
+  when: (item.value.hash | default('')) == ''
```
arjuhe, line 2: Glad this is gone.

arjuhe (approved):
Looks much better to me.

arjuhe (approved):
Looks better to me.

PR #187, opened 2022-11-11
Fix issue with deployment plan template

nocturnalastro:
Jinja was complaining that there was an unexpected end block removing the else and switching the +s for -s seems to have removed the issue


arjuhe (approved):
+1

PR #188, opened 2022-11-11
Add the ability to specify the olm operators that should be installed…

nocturnalastro:
… with the cluster

This makes use of a feature in assisted installer which allows the user to specify some specific operators to be installed when the cluster is created


arjuhe (approved):
LGTM!

PR #189, opened 2022-11-11
[WIP] Split vaildation tasks into roles

nocturnalastro:
This is so that vaildaions can be applied more graularly so that playbooks can be used easier on an individual basis


arjuhe:
Closing pull request as project is being archived.

PR #190, opened 2022-11-17
Allow users to specify node labels to be applied post install

nocturnalastro:


nocturnalastro:
This is part of the gitops hook PR.

PR #192, opened 2022-11-23
Have sushy tools installed in a virtual env

thekad:
Currently, sushy tools is installed in the system-wide python path, this causes issues with e.g. newer versions of libraries in the sushy releases, and overwriting RPM files with PIP files. Also adjusting the systemd service file to use the virtual environment.

### DIFF
--- a/roles/setup_sushy_tools/defaults/main.yml
+++ b/roles/setup_sushy_tools/defaults/main.yml
@@ -8,6 +8,10 @@ sushy_data_dir: "{{ sushy_dir }}/data"
 sushy_packages:
   - python3-pip
   - python3-virtualenv
+  - libvirt-devel
+sushy_pip_packages:
+  - sushy-tools
+  - libvirt-python
 
 file_owner: "{{ ansible_env.USER }}"
 file_group: "{{ file_owner }}"
--- a/roles/setup_sushy_tools/defaults/main.yml
+++ b/roles/setup_sushy_tools/defaults/main.yml
@@ -8,6 +8,7 @@ sushy_data_dir: "{{ sushy_dir }}/data"
 sushy_packages:
   - python3-pip
   - python3-virtualenv
+  - gcc
   - libvirt-devel
 sushy_pip_packages:
   - sushy-tools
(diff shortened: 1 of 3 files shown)


thekad:
cc @p3ck

nocturnalastro:
I hit this error `sushy-emulator[900973]: [2022-11-25 05:32:14,748] ERROR in main: libvirt driver not loaded` I think the python package for the libvirt bindings isn't in the virtual env

nocturnalastro:
A quick attempt at adding libvirt-python failed.

thekad:
Hmmmm, my virtualenv must've been polluted, let me make an attempt from zero and will try to figure it out

thekad:
@nocturnalastro it wasn't my virtualenv, rather my machine had libvirt-devel already installed, can you take a look please ?

roles/setup_sushy_tools/tasks/main.yml
```diff
@@ -2,9 +2,27 @@
...
+        - "{{ sushy_auth_dir }}"
+        - "{{ sushy_cert_dir }}"
+        - "{{ sushy_data_dir }}"
+
+    - name: Install sushy-tools via pip in a virtual environment
       pip:
-        name: "sushy-tools"
+        name: "{{ item }}"
+        virtualenv: "{{ sushy_dir }}"
+      loop: "{{ sushy_pip_packages }}"
```
nocturnalastro, line 25: Name can take a list no need to loop.

roles/setup_sushy_tools/defaults/main.yml
```diff
@@ -5,6 +5,13 @@ sushy_auth_dir: "{{ sushy_dir }}/auth"
 sushy_cert_dir: "{{ sushy_dir }}/cert"
 sushy_auth_file: "{{ sushy_auth_dir }}/htpasswd"
 sushy_data_dir: "{{ sushy_dir }}/data"
+sushy_packages:
+  - python3-pip
+  - python3-virtualenv
+  - libvirt-devel
```
nocturnalastro, line 11: Looks like from my test that you also need `gcc` :)

nocturnalastro (changes requested):
After adding `gcc` my test passed sushy tool the sushy tool interactions. Once that is added and the loop is removed, it looks good to me :)

nocturnalastro (approved):
Thanks for this :) @arjuhe could you take a look at this :)

arjuhe (approved):
LGTM

PR #193, opened 2022-11-25
Only use schedulable_masters when cluster has workers

nocturnalastro:
Without workers masters have to schedulable so we will ignore the user provided parameter. Also removed manifest as it meant that the user var meant nothing as masters will always be schedulable regardless of the user's choice of value.


arjuhe (approved):
LGTM

PR #196, opened 2022-12-05
Adding worker node with Multi Cluster Engine

sarpkoksal:
Adding a functionality that adding a worker node with Multi Cluster Engine based on AI Team's feedback

### DIFF
--- a/playbooks/approve_add_host.yml
+++ b/playbooks/approve_add_host.yml
@@ -1,0 +1,6 @@
+- hosts: bastion
+  roles:
+    - role: approve_add_host
+      when: groups['day2_workers'] | default([]) | length > 0
+  vars:
+    - disconnected: "{{ use_local_mirror_registry | default(setup_registry_service | default(true)) }}"
--- a/roles/apply_nmstate/tasks/main.yml
+++ b/roles/apply_nmstate/tasks/main.yml
@@ -7,11 +7,11 @@
         name: nmstate
         state: present
 
-    - name: "Copy rendered_nmstate_yml to {{ vm_nmstate_config_path }}"
-      copy:
-        content: "{{ rendered_nmstate_yml }}"
-        dest: "{{ vm_nmstate_config_path }}"
-        mode: 0644
+   # - name: "Copy rendered_nmstate_yml to {{ vm_nmstate_config_path }}"
+   #   copy:
+   #     content: "{{ rendered_nmstate_yml }}"
+   #     dest: "{{ vm_nmstate_config_path }}"
+   #     mode: 0644
       # No commit is done to revert the changes if they cause the host to be come unreachable
 
     - name: Apply nmstate
--- a/roles/approve_add_host/tasks/main.yml
+++ b/roles/approve_add_host/tasks/main.yml
@@ -1,0 +1,16 @@
+- name: Get Agent name
+  shell: "oc get agent -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} -o json | jq -r .items[].metadata.name"
+  register: oc_get_agent_name
+
+- name: Approve the host
+  shell: "oc patch agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} --type merge --patch '{\"spec\":{\"approved\":true }}'"
+
+- name: Add host to cluster
+  shell: "oc patch agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} --type merge --patch '{\"spec\":{\"clusterDeploymentName\":{\"name\": \"{{ namespace }}\", \"namespace\": \"{{ namespace }}\"}}}'"
+
+- name: Wait until Host is installed
+  command: "oc get agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} -o json"
+  register: oc_get_managedcluster
+  until: oc_get_managedcluster.stdout|from_json|json_query('status.conditions[*].status')|unique == ["True"]
+  retries: 20
+  delay: 60
--- a/roles/create_mce_crs/files/07_create-cluster-deployment-template.yml
+++ b/roles/create_mce_crs/files/07_create-cluster-deployment-template.yml
@@ -1,0 +1,57 @@
+---
+apiVersion: v1
+kind: Template
+labels:
+  template: create-clusterdeployment-template
+metadata:
+  name: create-clusterdeployment-template
+objects:
+- apiVersion: hive.openshift.io/v1
+  kind: ClusterDeployment
+  metadata:
+    name: ${NAMESPACE}
+    namespace: ${NAMESPACE}
+  spec:
+    baseDomain: ${BASEDOMAIN}
+    clusterInstallRef:
+      group: extensions.hive.openshift.io
+      kind: AgentClusterInstall
+      name: ${NAMESPACE}
+      version: v1beta1
+    clusterMetadata:
+      adminKubeconfigSecretRef:
+        name: ${NAMESPACE}-admin-kubeconfig
+      adminPasswordSecretRef:
+        name: ${NAMESPACE}-admin-password
+      clusterID: "${CLUSTERID}"
+      infraID: "${INFRAID}"
+    clusterName: ${CLUSTERNAME}
+    installed: true
+    platform:
+      agentBareMetal:
+        agentSelector:
+          matchLabels:
+            cluster-name: ${CLUSTERNAME}
+    pullSecretRef:
+      name: pull-secret
+parameters:
+- description: The namespace that will be created for imported cluster
+  name: NAMESPACE
+  value: "noknom-aicli"
+  required: true
+- description: The Base domain
+  name: BASEDOMAIN
+  value: "cloud.lab.eng.bos.redhat.com"
+  required: true
+- description: The OCP version
+  name: CLUSTERID
+  value: "8581800e-e3e6-41ae-a391-28c95e9166b1"
+  required: true
+- description: The sshkey
+  name: INFRAID
+  value: "crucible-fhcpd"
+  required: true
+- description: The sshkey
+  name: CLUSTERNAME
+  value: "crucible"
+  required: true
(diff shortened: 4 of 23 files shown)


roles/apply_nmstate/tasks/main.yml
```diff
@@ -7,11 +7,11 @@
         name: nmstate
         state: present
 
-    - name: "Copy rendered_nmstate_yml to {{ vm_nmstate_config_path }}"
-      copy:
-        content: "{{ rendered_nmstate_yml }}"
-        dest: "{{ vm_nmstate_config_path }}"
-        mode: 0644
+   # - name: "Copy rendered_nmstate_yml to {{ vm_nmstate_config_path }}"
```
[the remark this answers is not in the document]
sarpkoksal: I had to comment it out because it was giving me error while appyling the nmstate yml in the kvm host. Error was about dns-resolver thing. You might remember it from previous cases as well couple of months ago. I had to run the nmstate yml manually so I commented out this part in the roles but I can remove the comments from this part.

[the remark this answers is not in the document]
nocturnalastro: If you don't want crucible to apply the nmstate to the vmhost you can add the flag 
`SETUP_VM_BRIDGE: False` to your inventory. :) Saving you having to comment it out :)

roles/create_mce_crs/files/07_create-cluster-deployment-template.yml
```diff
@@ -0,0 +1,57 @@
+---
+apiVersion: v1
+kind: Template
```
nocturnalastro, line 3: Is there a need to use the k8s templates instead of using Jinja? This adds another form of template into the codebase which seems unnecessary. 

You are even passing the values to the templates from ansible so why not use jinja?

Which also seems to lead a lot of tasks adding these templates to the cluster.

roles/generate_discovery_iso_mce/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+- name: Get CA Secret and save to ca-trust path
+  shell: "oc get secret router-ca -o json -n openshift-ingress-operator |jq -r \".data.\\\"tls.crt\\\"\"|base64 -d > /etc/pki/ca-trust/source/anchors/nokianom.crt "
+  register: cert
+
+- name: Update CA Trust
```
nocturnalastro, line 5: This needs `become: true`

playbooks/approve_add_host.yml
```diff
@@ -0,0 +1,6 @@
+- hosts: bastion
+  roles:
+    - role: approve_add_host
+      when: groups['day2_workers'] | default([]) | length > 0
+  vars:
+    - disconnected: "{{ use_local_mirror_registry | default(setup_registry_service | default(true)) }}"
```
nocturnalastro, line 6: Where is this used?

roles/create_mce_crs/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+- name: Create Cluster Image Set
+  shell: "oc process create-cluster-image-set-template -p OCPVERSION={{ openshift_full_version }} | oc create -f -"
+
+- name: Create Agent Cluster Install
+  shell: "oc process create-agentclusterinstall-template -p NAMESPACE={{ namespace }} -p APIVIP='{{ api_vip }}' -p OCPVERSION='{{ openshift_full_version }}' -p INGRESSVIP='{{ ingress_vip }}' -p PLATFORMTYPE={{ platform_type }} -p CLUSTERNETWORK={{ cluster_network_cidr }} -p SERVICENETWORK={{ service_network_cidr }} -p SSHKEY='{{ sshkey }}'| oc create -f -"
+
+- name: Create Kubeconfig
+  command: "oc -n {{ namespace }} create secret generic {{ namespace }}-admin-kubeconfig --from-file=kubeconfig={{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}}"
+
+- name: Get credentials
```
nocturnalastro, line 40: These creds are already available in a vault file. Although I suppose it doesn't hurt to retrieve them again, if the this is part of the same flow. Depending how long after the install this is run those creds might have been garbage collected.

roles/create_mce_crs/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+  command: "oc apply -f {{ repo_root_path }}/roles/create_mce_crs/files/05_create-agent-cluster-install-template.yml --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}}"
+
+- name: Create Template for Import Cluster (ClusterDeployment)
+  command: "oc apply -f {{ repo_root_path }}/roles/create_mce_crs/files/07_create-cluster-deployment-template.yml --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}}"
+
+- name: Create Template for Managed Cluster
+  command: "oc apply -f {{ repo_root_path }}/roles/create_mce_crs/files/managed-cluster-template.yml --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}}"
+
+- name: Create Namespace for imported cluster
+  shell: "oc process create-ns-template -p NAMESPACE={{ namespace }} | oc create -f -"
```
nocturnalastro, line 23: I don't see where the value for `namespace` is defined. If its user input then can you add a check for it.

roles/approve_add_host/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: Approve the host
+  shell: "oc patch agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} --type merge --patch '{\"spec\":{\"approved\":true }}'"
+
+- name: Add host to cluster
+  shell: "oc patch agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} --type merge --patch '{\"spec\":{\"clusterDeploymentName\":{\"name\": \"{{ namespace }}\", \"namespace\": \"{{ namespace }}\"}}}'"
+
+- name: Wait until Host is installed
+  command: "oc get agent {{ oc_get_agent_name.stdout }} -n {{ namespace }} --kubeconfig {{kubeconfig_dest_dir}}{{kubeconfig_dest_filename}} -o json"
+  register: oc_get_managedcluster
+  until: oc_get_managedcluster.stdout|from_json|json_query('status.conditions[*].status')|unique == ["True"]
```
nocturnalastro, line 14: Lets work together on this after the new year I think we can refactor this to use the kubenetes module, less shell and command calls is a much better way of doing ansible. 

Previously we used oc as it was only a few commands and we didn't see the need to add a new dependency. I think we have far gone passed the threshold where a new dependency is worth it :)

nocturnalastro (changes requested):
Lets work together on this after the new year I think we can refactor this to use the kubenetes module, less shell and command calls is a much better way of doing ansible.

Previously we used oc as it was only a few commands and we didn't see the need to add a new dependency. I think we have far gone passed the threshold where a new dependency is worth it :)

roles/generate_discovery_iso_mce/tasks/main.yml
sarpkoksal: I couldnt understand this one.

roles/create_mce_crs/files/07_create-cluster-deployment-template.yml
sarpkoksal: I wanted use templates because some of these custom resources that will be deployed requires parameters so I found this way but if Jinja is easier way, we can take a look at it. But this template method was not that hard and not lead to lot of tasks, it was easy :)

playbooks/approve_add_host.yml
sarpkoksal: This is used for approving and adding host to cluster after node is booted with iso and installed

roles/create_mce_crs/tasks/main.yml
sarpkoksal: It is an user input in "inventory.yml". What kind of check you mean?

sarpkoksal:
Ok we can work on this together after new year. I will take a look at this kubernetes module.

roles/create_mce_crs/tasks/main.yml
nocturnalastro: Lets take a multi step approach to this:

1. Check for a variable `day2_kubeamin_password` or something to that effect
2. Check for the file created in post_install
3. Attempt to retrieve the value from the assisted installer API
4. Fail and request the user set the variable.

roles/generate_discovery_iso_mce/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  shell: oc get infraenv -A -o json|jq -r '.items[0].status.isoDownloadURL'
+  register: iso_url
+
+- set_fact:
+      iso_url_str: "{{ iso_url.stdout | replace ('minimal' , 'full') }}"
+
+- name: Download ISO
+  get_url:
+    url: "{{iso_url_str}}"
+    dest: /opt/http_store/data/discovery/discovery_mce.iso
```
nocturnalastro, line 18: Construct the save path out of these vars that way if the http_store is not the crucible one the use can still modify the location.
```
DOWNLOAD_DEST_FILE: "{{ day2_discovery_iso_name }}"
DOWNLOAD_DEST_PATH: "{{ iso_download_dest_path | default('/opt/http_store/data') }}"
```

playbooks/approve_add_host.yml
sarpkoksal: deleted

roles/create_mce_operator/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+version: false
+disconnected: False
+manifests: True
+
+mirror_registry: "{{ hostvars['registry_host']['registry_fqdn'] }}:{{ hostvars['registry_host']['registry_port'] }}"
+
```
nocturnalastro, line 6: I think these can be deleted

roles/create_mce_crs/tasks/main.yml
nocturnalastro: Add a check to validate inventory perhaps with a var `use_mce_for_day2` and then validate_inventory to the meta of the role.

roles/generate_discovery_iso_mce/tasks/main.yml
sarpkoksal: Done

roles/apply_nmstate/tasks/main.yml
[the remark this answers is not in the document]
sarpkoksal: Done

roles/create_mce_operator/defaults/main.yml
sarpkoksal: Done

roles/generate_discovery_iso_mce/tasks/main.yml
sarpkoksal: done

nocturnalastro:
Superseded by #210

PR #197, opened 2022-12-07
Fix installation fail msg in monitor_cluster role

abays:
It looks like the cluster installation failure message for an encountered error is copy-pasted improperly (it indicates `user cancelled` when the task itself is checking for a cluster error)


nocturnalastro (approved):

arjuhe (approved):
+1

PR #198, opened 2022-12-12
Don't change kvm_node_name

p3ck:
Use the name given.

In our Telco CI lab we share vm resources between IPI and Assisted installs.  When assisted installer runs we tear down the VM's that were possibly created by the IPI install and vice versa.  The problem is crucible forces us to use the cluster name prepended to the vm's and this doesn't match what is used in the IPI installs.

Can I ask what the reason is for forcing the VM's names to be pre-pended by the cluster name?


nocturnalastro:
The reason was to prevent name collisions when using crucible to deploy multiple clusters on to the same host. I would like to keep this behaviour, in case people are relying on it without explicitly knowing it. Perhaps we could add a flag, to stop this but leave it on by default.

p3ck:
Updated so the default doesn't change the existing behavior.

roles/destroy_vms/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 - name: Create inventory host to vm name mapper
   set_fact:
-    cluster_vm_names: "{{ (cluster_vm_names | default([])) + [item.startswith(cluster_name) | ternary(item, (cluster_name + '_' + item))] }}"
```
nocturnalastro, line 3: I think the `item.startswith(cluster_name)` should be `item.startswith(vm_mode_prefix)`. I checked `'a random string'.startswith('') == True`

p3ck: Thanks for doing the test for me..  I did think about doing exactly that..  but I didn't think it would match for ''

I'll update the PR

p3ck:
Test passed in bos2 CI.

nocturnalastro (approved):
@arjuhe Can you take a look at this for me :)

arjuhe (approved):
lgtm

PR #204, opened 2023-01-05
Add version infomation to image_hashes file

nocturnalastro:
This is to stop issues when the image implicitly changes e.g. assisted installer version is bumped. Before this the user would have to know that they need to remove the image hashes file to get the new version.


arjuhe:
Closing pull request as project is being archived.

PR #205, opened 2023-01-05
Remove explicit boolean checks

nocturnalastro:


nocturnalastro:
Clean up does this and more

PR #206, opened 2023-01-05
Fix requirement for a predefined ssh key

nocturnalastro:


prereq_facts_check.yml
```diff
@@ -7,5 +7,5 @@
   roles:
     - prereq_facts_check
   vars:
-    ssh_public_check: "{{ not (generate_ssh_keys | default(False)) }}"
+    ssh_public_check: "{{ not (generate_ssh_keys | default(True)) }}"
```
arjuhe, line 10: Would removing the not do the same thing?

```suggestion
    ssh_public_check: "{{ generate_ssh_keys | default(Falee) }}"
```

arjuhe (changes requested):

prereq_facts_check.yml
nocturnalastro: No it needs to that way round. We only want this to run if user sets generate_ssh_keys to false we could do. We want to apply the `not` to `generate_ssh_keys` but by default and assumed to be true.

arjuhe (approved):

PR #207, opened 2023-01-06
Create day2 KVM nodes as part of day2 process

nocturnalastro:
This is becuase the day2 nodes maybe be added as a secondary run we must filter the kvm nodes otherwise the playbooks will destroy the current cluter.


arjuhe (approved):
Lgtm

arjuhe (changes requested):
Just a small change nothing major. (👍

arjuhe (approved):

PR #208, opened 2023-01-12
Add support to OCP nightly builds for populate_mirror_registry/prerequisites

ramperher:
This change started because it was needed to support OCP nightly builds, where `oc` client need to include the version before `.tar.gz`. Else, tests with assisted installer using nightly builds will fail, like in [this example](https://www.distributed-ci.io/jobs/9ae8ead2-2c6b-474b-a588-ed2c6f10cb87/jobStates?sort=date&task=e4753813-aa50-4dca-a107-fe365400e06a) that we have run in Telco CI labs.

The problem is, when having a nightly build, `release_url` is different (e.g. https://openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com/), and the way of interacting with these resources are a bit tricky, so they need to be handled in a different way.

In our dci-openshift-agent, we have already validated in [this change](https://softwarefactory-project.io/r/c/dci-openshift-agent/+/27122) that we can extract all the artifacts (in this case, the release.txt file and the oc binary) from the OCP release image, so release_url would no longer be needed.

Note that, for `opm` client, nightly builds are not currently including it, so this change also modifies it to rely on stable clients from mirror.openshift.com, which is what is being done also in dci-openshift-agent.

So, tldr, the change implies the following:
- Rely on OCP release image instead of the release URL to download the release.txt and the oc binary
- For this, a stable oc binary needs to be downloaded to extract the artifacts from the release image
- Now the release.txt is directly retrieved from the artifact, not from the release URL
- Now oc binary is directly unarchived from the artifact, not from the release URL
- opm client needs to be downloaded from stable release URL

### DIFF
--- a/roles/populate_mirror_registry/defaults/main.yml
+++ b/roles/populate_mirror_registry/defaults/main.yml
@@ -17,9 +17,13 @@ downloads_path: "{{ ansible_env.HOME }}"
 image_name_registry: ocpdiscon-registry
 
 # oc and opm install
-release_url: "https://mirror.openshift.com/pub/openshift-v{{ openshift_version_parts[0] }}/clients/ocp"
-oc_tar: openshift-client-linux.tar.gz
-opm_tar: opm-linux.tar.gz
+# - oc tar will be retrieved from the release image, needs OCP full version.
+#   Stable client is also used to extract the artifacts from release image
+# - opm tar always has to come from stable release URL, so that its name is well known
+stable_release_url: "https://mirror.openshift.com/pub/openshift-v{{ openshift_version_parts[0] }}/clients/ocp/stable"
+oc_tar: "openshift-client-linux-{{ openshift_full_version }}.tar.gz"
+stable_oc_tar: "openshift-client-linux.tar.gz"
+stable_opm_tar: "opm-linux.tar.gz"
 
 # The information for the locally created registry
 registry_fqdn: "{{ ansible_fqdn }}"
--- a/roles/populate_mirror_registry/tasks/prerequisites.yml
+++ b/roles/populate_mirror_registry/tasks/prerequisites.yml
@@ -1,4 +1,16 @@
 ---
+- name: Check that release_url variable is not defined
+  assert:
+    that:
+      - release_url is not defined
+    fail_msg: >
+      release_url is no longer needed in this role, the
+      functionality around it has changed.
+      Now, artifacts are directly extracted from release image,
+      and they are not downloaded from release_url.
+      Please make sure not to use release_url variable and
+      to check that you are using the correct release image
+
 - name: Make sure needed packages are installed
   package:
     name: tar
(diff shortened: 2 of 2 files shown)


ramperher:
Validated that the change is working fine in these two cases:

- If artifacts are present, do not download them: you can see in [this job](https://www.distributed-ci.io/jobs/b3fa3d55-7d45-4680-a3f7-aee8794d06ac/jobStates?sort=date&task=2a38ff8b-9dd8-43d1-8cab-11e228322c70) that these tasks are skipped.
- If artifacts are not present, download them: in [this job](https://www.distributed-ci.io/jobs/b9341183-88fb-4f9e-8211-c723d560a94f/jobStates?sort=date&task=faab279f-eb22-424f-a4ed-eb87a630298d), all the previously skipped tasks are now executed (the job failed due to other reasons, but the tasks related to this change are working).

roles/populate_mirror_registry/tasks/prerequisites.yml
```diff
@@ -16,39 +16,73 @@
...
+
+    - name: Extract artifacts from release image (arm64)
+      command:
+        cmd: >
+          "{{ oc_tmp_dir.path }}/oc" release extract
+            --registry-config "{{ config_file_path }}/{{ pull_secret_file_name }}"
+            --tools
+            --from "{{ release_image_item.remote | quote }}"
+            --to "{{ downloads_path }}/{{ openshift_full_version }}"
+      when: arm_release_image_key in image_hashes
```
nocturnalastro, line 63: Can you explain what this task is meant to do, it looks identical other than `adm` missing from  `/oc release extract`.

roles/populate_mirror_registry/defaults/main.yml
```diff
@@ -17,9 +17,10 @@ downloads_path: "{{ ansible_env.HOME }}"
 image_name_registry: ocpdiscon-registry
 
 # oc and opm install
-release_url: "https://mirror.openshift.com/pub/openshift-v{{ openshift_version_parts[0] }}/clients/ocp"
-oc_tar: openshift-client-linux.tar.gz
-opm_tar: opm-linux.tar.gz
+stable_release_url: "https://mirror.openshift.com/pub/openshift-v{{ openshift_version_parts[0] }}/clients/ocp/stable"
```
nocturnalastro, line 20: Could you add a check into validate inventory like we did [here](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/validate_inventory/tasks/vms.yml#L9-L17), that explains that `release_url` can no longer be used and the functionality around it has changed. As anyone overriding this would not get what they were expecting.

I don't think this neatly fits into any of the existing check files feel free to make a new one.

nocturnalastro (changes requested):
@ramperher Thanks for the PR using the release image is something we've been thinking of doing for a while but just never gotten round to do. 

Please see the comments above :)

roles/populate_mirror_registry/tasks/prerequisites.yml
ramperher: Oh, you're right, I'm missing `adm`, I'll add it.
The command is working because I've tested it in x86 (the task before), by the way.
the explanation of the task is above the block that includes all this logic, mainly this:
```
# extract artifacts from the OCP release image instead of downloading them
# from the release URL
# This OCP release image provides the artifacts required (OCP clients, installer,
# release.txt, among others)
```
I hope this is fine, else I can write more comments, but it's mainly the same that we've included for our dci-openshift-agent.

roles/populate_mirror_registry/defaults/main.yml
ramperher: Done, I hope it's fine now too.

ramperher:
> @ramperher Thanks for the PR using the release image is something we've been thinking of doing for a while but just never gotten round to do.
> 
> Please see the comments above :)

Thanks for your review @nocturnalastro! Yes, this method is a bit tricky, but after dealing with nightly builds in our labs, we have moved to this approach as it considerably more stable and without depending on fetching any external URL.
I've addressed all your comments, if you need me to check something more, don't hesitate to reach me.

roles/populate_mirror_registry/tasks/prerequisites.yml
```diff
@@ -16,39 +29,73 @@
...
+      command:
+        cmd: >
+          "{{ oc_tmp_dir.path }}/oc" adm release extract
+            --registry-config "{{ config_file_path }}/{{ pull_secret_file_name }}"
+            --tools
+            --from "{{ release_image_item.remote | quote }}"
+            --to "{{ downloads_path }}/{{ openshift_full_version }}"
+      when: arm_release_image_key not in image_hashes
+
+    - name: Extract artifacts from release image (arm64)
```
nocturnalastro, line 68: @ramperher Thanks for the explanation and the corrections. I'd think you would only want extract those artefacts if the bastion was arm? My reasoning behind this is that they are run on the bastion so the binaries should be for the bastion's architecture.

I think there are two options here: 

Either you can have var (something like `tool_extract_image`) which defaults to x86_64 release and document that if the bastion is arm they should change the release image and `stable_opm_tar`. 

Or check the ansible facts what the arch of the bastion is and use the correct release image and `stable_opm_tar` accordingly.

nocturnalastro (changes requested):
Once we have this selection of the arm stuff wrapped up could you please squash your commits :)

roles/populate_mirror_registry/tasks/prerequisites.yml
ramperher: Yes, that's it, I was trying to do something similar to what it is done [here](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/populate_mirror_registry/tasks/populate_registry.yml#L53),  where the same check is done. But effectively, this is just to separate the cases of x86_64 and arm in two different tasks.

In our case, as we're always using x86_64, we were not paying attention to arm64. But now, I can see your point, and also that the `stable_oc/opm_tar` should point to arm one in case of being on arm64.

I'm going to follow your first approach, because in our case, we're not gathering Ansible facts by default, so this may affect to the performance of our playbooks. I'll rename the variable to `release_arch`, just to have a meaningful variable name, and I will refactor the `stable_oc/opm_tar` and the Extract artifacts tasks to simplify the logic a little bit more.

I'll also squash the commits after this, thanks!

ramperher: Now I've realized something: `oc_tar` changes depending on the architecture, you can see this in examples [(1)](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/) for a stable version and [(2)](https://openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com/4.13.0-0.nightly-2023-01-11-225449/) for a nightly version. But in the case of opm, where we always have to go to the stable repository (it is not available in the nightly repo), we don't have a distinction between x86_64 and arm64, so there, we'll directly download the only binary available.

ramperher:
I've done some refactoring work to make some failing tasks to work. You can find them here:

- Case where we've already downloaded all the software, skipping tasks is working - starting from [here](https://www.distributed-ci.io/jobs/c895b63e-abdf-459d-bfc3-2d730973d12b/jobStates?sort=date&task=b65ca44f-e9a3-41cf-a2aa-f4f8ed42c70a)
- Case where we have to download everything - starting from [here](https://www.distributed-ci.io/jobs/0555c1fa-4f55-42d0-a22f-27d9ec459d55/jobStates?sort=date&task=c3bc316e-f0fd-4d5d-8caf-71f3caebc157)

roles/populate_mirror_registry/tasks/prerequisites.yml
nocturnalastro: You can probably use `ansible_architecture`

nocturnalastro: Ah I didn't realise that about OPM. Given that the tar has not arch in the name and without ensuring/testing that OPM works on an arm64 machine. I think the best route is to assume that the machine is `x86_64` as the previous code did. And remove the arm stuff entirely.


The issue I'm trying to avoid is it looking like it should work for arm and then blowing up because OPM doesn't work. It also simplifies the code which is nice.

nocturnalastro (changes requested):
Thanks for all your work so far. Sorry to keep making you change things, I'm just trying to make sure we get the best PR possible and as new information comes to light the definition of best keeps changing.

ramperher:
> Thanks for all your work so far. Sorry to keep making you change things, I'm just trying to make sure we get the best PR possible and as new information comes to light the definition of best keeps changing.

Sure, no worries, I've reupdated everything to avoid checking arm case, I hope it's fine now

nocturnalastro (approved):
@arjuhe Could you take a look at this please :)

arjuhe (approved):
LGTM

PR #209, opened 2023-01-12
Clean up

nocturnalastro:


nocturnalastro:
@arjuhe Only a few errors left and fixing them makes things worse not better :)

arjuhe:
Doesn't it sting when fixing things makes it worse. We will need to do a synchronous review for this one.

nocturnalastro:
This is too big to practically review and now its out of date. I'll split this into smaller cunks instead.

PR #214, opened 2023-01-31
Added a small playbook to create the crucible user if needed.

arjuhe:


nocturnalastro (approved):

natifridman:
Need to also add the key to [authorized keys](https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html)

arjuhe:
This poses a problem for anyone who is using a pre-exisiting assisted-installer. It appears that the DB is not compatible with the new version. I had to destroy all previous versions of assisted-installer and do a clean deploy for this to work.

PR #215, opened 2023-02-02
Vendor PXE it's possible the IPMI port could be on a non-standard port

p3ck:
It's possible that the IPMI service is not on the default port of 623.  Specifying bmc_port in inventory allows you to override the default.

I'm open to suggestions.  The original example in docs/inventory.yml (which doesn't work) has the port in the address.  An alternate solution would be to look for : in bmc_address and pull the port out of it if it exists.  If we go that route we will need to update the validate network role where it tries to ping bmc_address.


nocturnalastro (approved):

arjuhe (approved):
LGTM

PR #216, opened 2023-02-07
Make kubeadmin creds output in nice json

nocturnalastro:
This allows for jq to read the creds easily making it easer to work with the kubeadmin vault


arjuhe (approved):
What's wrong with yaml, not pretty enough?

nocturnalastro:
jq doesn't read yaml :) and yaml tools can read json :)

PR #218, opened 2023-02-08
Bump assisted installer version to v2.12.1

nocturnalastro:
This is the value being used for ai_version by the DCI team


arjuhe:
This poses a problem for anyone who is using a pre-exisiting assisted-installer. It appears that the DB is not compatible with the new version. I had to destroy all previous versions of assisted-installer and do a clean deploy for this to work.

nocturnalastro:
> I ran this in our lab against a pre-existing assisted installer and it remained v2.1.1.

This turned out to be because the `image-hashes.yml` file PR #204 should prevent this issue once merged.

arjuhe (approved):

PR #220, opened 2023-02-14
Fix support for RHEL 9

nocturnalastro:
The changes requred to support RHEL 9 and ansible-core 2.11 where:
- use full name for ansible.utils.hwaddr
- use full name for ansible.utils.ipaddr
- update requirements.yml adding ansible.utils

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,6 +1,7 @@
 ---
 collections:
   - ansible.posix
+  - ansible.utils
   - containers.podman
   - community.crypto
   - name: community.general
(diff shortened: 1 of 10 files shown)


nocturnalastro:
Looks like I need to make the tests flow install the galaxy collections

nocturnalastro:
Tested for backwards compatibility with RHEL8.7 in lab

requirements.yml
```diff
@@ -1,6 +1,7 @@
 ---
 collections:
   - ansible.posix
+  - ansible.utils
```
arjuhe, line 4: Does this require a psa to alert the community to reinstall ansible modules?

arjuhe (approved):

requirements.yml
nocturnalastro: Yes probably :)

nocturnalastro: I still wish we had done my idea for pre-run messages would make these things easer.

PR #222, opened 2023-02-21
Remove temp ssh key dir after moving to perminant storage. 

nocturnalastro:


roles/generate_ssh_key_pair/tasks/main.yml
```diff
@@ -52,6 +52,11 @@
 
   delegate_to: bastion
 
+- name: "Remove temporary ssh_keys folder {{ key_pair_dir }}"
+  file:
+    path: "{{ key_pair_dir }}"
+    state: absent
+
```
arjuhe, line 59: I am splitting hairs here but wdyt about renaming {{ key_pair_dir }} to {{ tmp_keypair_dir }}?

nocturnalastro: We can do that but it could break people's inventories if they are overriding it.

arjuhe:
damn inventories.

PR #225, opened 2023-02-27
Fix the check so it doesn't require uppercase hex

p3ck:
libvirt doesn't require the mac address to be upper case.   This changes the validation to accept both upper and lower.

### DIFF
--- a/roles/validate_inventory/tasks/cluster.yml
+++ b/roles/validate_inventory/tasks/cluster.yml
@@ -78,10 +78,10 @@
 - name: Assert mac has linux format
   assert:
     that:
-      - ( hostvars[item]['mac'] | string | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
+      - ( hostvars[item]['mac'] | string | upper | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
     quiet: true
     fail_msg: |-
-      The mac address for node {{ item }} needs to be in linx format XX:XX:XX:XX:XX:XX
+      The mac address for node {{ item }} needs to be in linux format XX:XX:XX:XX:XX:XX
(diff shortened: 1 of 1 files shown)


roles/validate_inventory/tasks/cluster.yml
```diff
@@ -78,7 +78,7 @@
 - name: Assert mac has linux format
   assert:
     that:
-      - ( hostvars[item]['mac'] | string | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
+      - ( hostvars[item]['mac'] | string | upper | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
     quiet: true
     fail_msg: |-
       The mac address for node {{ item }} needs to be in linx format XX:XX:XX:XX:XX:XX
```
arjuhe, line 84: Small typo on this line `linx`.

roles/validate_inventory/tasks/cluster.yml
```diff
@@ -78,7 +78,7 @@
 - name: Assert mac has linux format
   assert:
     that:
-      - ( hostvars[item]['mac'] | string | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
+      - ( hostvars[item]['mac'] | string | upper | regex_search('^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$')) is not none
```
arjuhe, line 81: 👍

arjuhe (changes requested):

roles/validate_inventory/tasks/cluster.yml
p3ck: Thanks, fixed that as well

nocturnalastro (approved):

PR #229, opened 2023-03-15
Fix issues with dns validation

nocturnalastro:
With some record configurations dig will return more than one line which broke the previous approch. Is stead we now check for an expected value being present in the returned string.

### DIFF
--- a/roles/validate_dns_records/defaults/main.yml
+++ b/roles/validate_dns_records/defaults/main.yml
@@ -1,7 +1,13 @@
 required_domains:
-  - "{{ 'api.' + domain }}"
-  - "{{ 'api-int.' + domain }}"
-  - "{{ '*.apps.' + domain }}"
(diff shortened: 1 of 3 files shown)


roles/validate_dns_records/defaults/main.yml
```diff
@@ -1,7 +1,13 @@
...
+  "apps": "*.apps.{{ domain }}"
+
+expected_answers:
+  "api": "{{ api_vip }}"
+  "api-int": "{{ api_vip }}"
+  "apps": "{{ ingress_vip }}"
+
 required_binary: dig
 required_binary_provided_in_package: bind-utils
-domain: "{{ cluster_name }}.{{ base_dns_domain }}"
```
arjuhe, line 7: Bad spacing

arjuhe (changes requested):

arjuhe (approved):

PR #230, opened 2023-03-16
Fix dns check for ipv6 and network connection check for nmstate

nocturnalastro:


roles/apply_nmstate/tasks/main.yml
```diff
@@ -3,23 +3,36 @@
...
     - name: Apply nmstate
-      shell:
+      ansible.builtin.shell:
         cmd: "nmstatectl apply --no-commit --timeout 120 {{ vm_nmstate_config_path }}"
       async: 60
       poll: 5
 
+    - name: "Check for connection wider network"
+      ansible.builtin.shell:
+        cmd: "ping -c 4 {{ vm_network_test_ip }}"
```
arjuhe, line 29: This ping will default from the interface with the default route. Also, I believe the ping default timeout is 4secs. You might want to bring the timeout down to 1sec.

arjuhe (approved):

roles/apply_nmstate/tasks/main.yml
nocturnalastro: I think default route is fine as it's just to check we still can connect in if need be. I've added the 1s timeout though

thekad:
@nocturnalastro this PR introduces a new variable `vm_network_test_ip` which is not mentioned anywhere else, no idea how/where it's set up

nocturnalastro:
@thekad This should be provided by the user in there inventory. I'll have to figure up how to do a decent validation for this value but it requires connecting to the machine to check if the `vm_host` or the bastion is the same machine.

It should be something like the network gateway. The idea is to stop any `vm_host` `network_config` from taking down access to the machine. The previous implementation worked as long as the `vm_host` and bastion were different machines however if not, it can still confirm a `network_config` where there is no network connectivity. 

We could try and get the value from `network_config` its self or use the gateway from the current connection but we need some form of lighthouse.

PR #234, opened 2023-04-18
Improvements found on KVM testing

thekad:
* Increasing the timeout when inserting the virtual media iso
* Correctly labeling the http content for SELinux
* Adding ability to mount host volumes on the assisted installer pods


p3ck (approved):
LGTM

PR #242, opened 2023-04-25
Default the qemu_user to ansible user

thekad:
We have found in our particular use case we don't want to login as the root user on the target VM Host but we also don't want the VMs to be created as some non-root user. This way we can set `ansible_user` to our non-root user and set `qemu_user` as root. The default (current) behavior stays the same.

### DIFF
--- a/roles/create_vms/tasks/main.yml
+++ b/roles/create_vms/tasks/main.yml
@@ -1,8 +1,4 @@
 ---
-- name: Set qemu user
-  set_fact: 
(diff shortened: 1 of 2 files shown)


roles/create_vms/tasks/main.yml
```diff
@@ -1,7 +1,8 @@
 ---
 - name: Set qemu user
-  set_fact: 
+  set_fact:
     qemu_user: "{{ ansible_user | default(ansible_env.USER) }}"
```
nocturnalastro, line 4: would it be better to just do this in the defaults?

PR #245, opened 2023-05-17
Added comments in inventory.md

drykxs:
About the NICs status and DNS server IPs. These are important to keep in mind mostly when setting up a baremetal environment where all the interfaces are connected to some switch.

### DIFF
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -191,10 +191,11 @@ network_config:
         slaves:
           - ens7f0
           - ens7f1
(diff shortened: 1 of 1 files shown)


drykxs:
I added two comments to the inventory about the NICs status and DNS server IPs. These are important to keep in mind mostly when setting up a baremetal environment where all the interfaces are connected to some switch.

nocturnalastro (approved):

PR #246, opened 2023-06-08
populate_mirror_registry | Use a tmp folder for oc/kubectl/opm binaries

ramperher:
- Right now, we are placing all the binaries in `/usr/local/bin`
- It may happen that, if we are sharing the jumphost between multiple OCP clusters, this file path is common for all the deployments.
- It is better to have a dedicated folder for each deployment to save the binaries and use them whenever they are needed.
- Doing this in the populate_mirror_registry.
- Apart from this, we keep the `oc` and `kubectl` binaries in `/usr/local/bin` directories, since they are used in other roles. It is not the case of `opm` binary.


nocturnalastro (approved):
Thanks for this :)

PR #252, opened 2023-07-13
Gather more Assisted logs

ramperher:
- If installation fails, try to run `openshift-install gather bootstrap` command to retrieve logs from the bootstrap node (master-0)
- Backup agent-config files

### DIFF
--- a/roles/generate_manifests/tasks/main.yml
+++ b/roles/generate_manifests/tasks/main.yml
@@ -38,6 +38,17 @@
     - agent-config.yaml.j2
     - install-config.yaml.j2
 
+# The agent-config files are eventually deleted after its usage, just
+# save them in a safe place in case of needing to check them
+- name: Save a backup of agent-config files
+  fetch:
+    src: "{{ manifests_dir }}/{{ item }}"
+    dest: "{{ fetched_dest }}/{{ item }}"
+    flat: true
+  loop:
+    - agent-config.yaml
+    - install-config.yaml
+
 - name: Create extra_manifest_dir dir
   ansible.builtin.file:
     name: "{{ extra_manifest_dir }}"
--- a/roles/monitor_agent_based_installer/tasks/main.yml
+++ b/roles/monitor_agent_based_installer/tasks/main.yml
@@ -4,7 +4,29 @@
     chdir: "{{ manifests_dir }}"
   ignore_errors: True # Timeout is fixed and some bare metal clusters complete bootstrap just after the timeout
 
-- name: Wait for install complete
-  ansible.builtin.shell:
-    cmd: "{{ agent_based_installer_path }} --log-level=debug agent wait-for install-complete"
-    chdir: "{{ manifests_dir }}"
+- name: Check installation and gather jobs if it fails
+  block:
+    - name: Wait for install complete
+      ansible.builtin.shell:
+        cmd: "{{ agent_based_installer_path }} --log-level=debug agent wait-for install-complete"
+        chdir: "{{ manifests_dir }}"
+
+  rescue:
+    # Using master-0 IP address to reach the bootstrap VM
+    # Placing the logs in repo_root_path
+    # Trying several times in case there are SSH connectivity issues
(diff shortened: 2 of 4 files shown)


roles/generate_manifests/tasks/main.yml
```diff
@@ -38,6 +38,16 @@
     - agent-config.yaml.j2
     - install-config.yaml.j2
 
+# The agent-config files are eventually deleted after its usage, just
+# copy them in a safe place in case of needing to check them
+- name: Save a backup of agent-config files
+  copy:
```
nocturnalastro, line 44: This should probably be a fetch task and put it into the fetched dir. Like [this task](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/generate_ssh_key_pair/tasks/main.yml#L28)

roles/monitor_agent_based_installer/tasks/main.yml
```diff
@@ -4,7 +4,31 @@
...
+        cmd: "{{ agent_based_installer_path }} --log-level=debug agent wait-for install-complete"
+        chdir: "{{ manifests_dir }}"
+
+  rescue:
+    # Using master-0 IP address to reach the bootstrap VM
+    # Placing the logs in repo_root_path
+    # Trying several times in case there are SSH connectivity issues
+    - name: Gather logs from installation
+      vars:
+        bootstrap_host: >-
```
nocturnalastro, line 20: The node should be selected like [this](https://github.com/redhat-partner-solutions/crucible/blob/35bebe48675ec73c0809338de4fa945bafc474bd/roles/generate_manifests/defaults/main.yml#L11) as that is how its specified as the [bootstrap IP](https://github.com/redhat-partner-solutions/crucible/blob/35bebe48675ec73c0809338de4fa945bafc474bd/roles/generate_manifests/templates/agent-config.yaml.j2#L4)

nocturnalastro (changes requested):
Just some minor tweaks for consistency

roles/monitor_agent_based_installer/tasks/main.yml
ramperher: Good to know it! Since it's defined in a different role, I'll create the same var for this task and just reuse it

nocturnalastro: Yes just make sure you using the first master instead of the first node. They might be the same but they don't have to be :)

roles/generate_manifests/tasks/main.yml
ramperher: updated

roles/generate_manifests/tasks/main.yml
```diff
@@ -38,6 +38,17 @@
     - agent-config.yaml.j2
     - install-config.yaml.j2
 
+# The agent-config files are eventually deleted after its usage, just
+# save them in a safe place in case of needing to check them
+- name: Save a backup of agent-config files
+  fetch:
+    src: "{{ item }}"
+    dest: "{{ repo_root_path }}"
```
nocturnalastro, line 46: I think this should be `fetched_dest` for consistency with all the other fetched artefacts :)

nocturnalastro (changes requested):
Just this one little change and I think we're good :)

roles/generate_manifests/tasks/main.yml
ramperher: Sure! done. I'll recheck this because I think the automation is not working fine on this, I'll let you know when it's stable

ramperher:
> Please squash your commits then I'll merge :)

Sure, let me check that the debug job is working and then I'll squash and ping you, just to confirm it's fine

ramperher:
> > Please squash your commits then I'll merge :)
> 
> Sure, let me check that the debug job is working and then I'll squash and ping you, just to confirm it's fine

done, I've added a last fix for the variables used to retrieve the rendezvouz IP and now it's fine, all commits squashed, so this is ready for merging

Issue #255, opened 2023-07-18
nmstate package installation missing in non-virtual deployments


manurodriguez:
### Bug description

I've observed the following error in Baremetal deployments, the task fails because it does not find `nmstatectl` command.
```
TASK [generate_agent_iso : Generate ISO] ***********************************************************
1s
task path: /var/lib/dci-pipeline/assisted_deploy_repo/roles/generate_agent_iso/tasks/main.yml:8
fatal: [bastion]: FAILED! => {"changed": true, "cmd": "/home/dciteam/clusterconfigs-cluster6/extract/openshift-install --log-level=debug agent create image", "delta": "0:00:00.060114", "end": "2023-07-12 14:39:39.950449", "msg": "non-zero return code", "rc": 1, "start": "2023-07-12 14:39:39.890335", "stderr": "level=debug msg=OpenShift Installer 4.14.0-0.nightly-2023-07-11-092038
level=debug msg=Built from commit a4604b060b2abb913926d0485760d069a95d157f
level=debug msg=Fetching Agent Installer ISO...
level=debug msg=Loading Agent Installer ISO...
level=debug msg=  Loading Agent Installer Artifacts...
...
level=debug msg=        Fetching NMState Config...
level=debug msg=          Fetching Agent Config...
level=debug msg=          Reusing previously-fetched Agent Config
level=debug msg=          Fetching Install Config...
level=debug msg=          Reusing previously-fetched Install Config
level=debug msg=        Generating NMState Config...
...
level=error msg=<nmstatectl gc> failed, errorCode -1, stderr exec: \"nmstatectl\": executable file not found in $PATH, input yaml <interfaces: pkg=manifests
```

Currently NMstate package is installed but only in [virtual environments](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/setup_vm_host_network/tasks/main.yml#L9-L11) with the [apply_nmstate role](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/apply_nmstate/tasks/main.yml#L5-L8)

Versions affected:
- Testing 4.12 and 4.14, but this will occur in any OCP version.

Steps to reproduce:
- Prepare an inventory for a baremetal deployment
- launch the installation with crucible
- In the node where assisted installer will be running no nmstate packages are installed
- deployment fails in the `generate_agent_iso : Generate ISO` task.

### OpenShift version

other (provide in the description)

### Assisted Installer version

other (provide in the description)

### Required statements

[...]

PR #256, opened 2023-07-19
make sure nmstate is installed before generating ISO

manurodriguez:
Fixes: #255

### DIFF
--- a/roles/generate_agent_iso/tasks/main.yml
+++ b/roles/generate_agent_iso/tasks/main.yml
@@ -5,12 +5,22 @@
     mode: 0755
     recurse: true
 
-- name: Generate ISO
-  ansible.builtin.shell:
-    cmd: "{{ agent_based_installer_path }} --log-level=debug agent create image"
-    chdir: "{{ manifests_dir }}"
-  environment:
-    XDG_RUNTIME_DIR: "{{ config_file_path }}"
+- name: Generate an ISO and print message if there is a failure
+  block:
+    - name: Generate ISO
+      ansible.builtin.shell:
+        cmd: "{{ agent_based_installer_path }} --log-level=debug agent create image"
+        chdir: "{{ manifests_dir }}"
+      environment:
+        XDG_RUNTIME_DIR: "{{ config_file_path }}"
+      register: gen_iso
+  rescue:
+    - name: Print message about nmstate package missing
+      fail:
+        msg: "nmstate package missing in the bastion. See Dependencies section in README.md for more details."
+      when:
+        - gen_iso.stderr is defined
+        - gen_iso.stderr | regex_search("nmstatectl(.*) executable file not found")
 
 - name: Put discovery iso in http store
   delegate_to: http_store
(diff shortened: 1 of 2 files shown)


manurodriguez:
In the test above, the task passed, it didn't install nmstate because this is a KVM lab and the install it's done in another task before, but now this fix #255 
```
TASK [generate_agent_iso : Install nmstate] ******************************************************** 4s
task path: /var/lib/dci-openshift-agent/gerrit/28885-0122dcecbb9c0e661db90bd65150cdb7/crucible/roles/generate_agent_iso/tasks/main.yml:1
ok: [bastion] => {"changed": false, "msg": "Nothing to do", "rc": 0, "results": []}
```

nocturnalastro (changes requested):
Ah sorry I forgot that the agent based installer stuff is all done on the bastion. We have a rule that we can not assume `sudo` privileges on the bastion. This is why we have other hosts for the services. Lots of people use a setup where the bastion doubles up as service host but we can not assume this.

I think the solution (or at least band-aid) to this issue is to add `nmstate` [here](https://github.com/redhat-partner-solutions/crucible/blob/main/README.md?plain=1#L64) and [here](https://github.com/redhat-partner-solutions/crucible/blob/main/README.md?plain=1#L71) such that the user can install themselves or ask the admin of the bastion to do it for them. 

Perhaps adding a rescue block around the install and then raise an error with fail would be a good way of allowing users with sudo access to reap the benefit of automation but allowing users without it to see a nice error message with something like install "this" package.

manurodriguez:
Thanks for the feedback

> I think the solution (or at least band-aid) to this issue is to add `nmstate` [here](https://github.com/redhat-partner-solutions/crucible/blob/main/README.md?plain=1#L64) and [here](https://github.com/redhat-partner-solutions/crucible/blob/main/README.md?plain=1#L71) such that the user can install themselves or ask the admin of the bastion to do it for them.
> 
Done

> Perhaps adding a rescue block around the install and then raise an error with fail would be a good way of allowing users with sudo access to reap the benefit of automation but allowing users without it to see a nice error message with something like install "this" package.
>
Done

roles/generate_agent_iso/tasks/main.yml
```diff
@@ -5,12 +5,22 @@
...
+      environment:
+        XDG_RUNTIME_DIR: "{{ config_file_path }}"
+      register: gen_iso
+  rescue:
+    - name: Print message about nmstate package missing
+      debug:
+        msg: "nmstate package missing in the bastion. See Dependencies section in README.md for more details."
+      when:
+        - gen_iso.stdout is defined
+        - gen_iso.stdout | regex_search("nmstatectl(.*) executable file not found"
```
nocturnalastro, line 23: It writes this out to stdout not stderr?

manurodriguez: good catch, fixed. I'll run a test that will fail in this task because a lack of nmstate pkg to validate the output

manurodriguez:
I squashed the commits and tested two validate the two scenarios:
Failed job because nmstate was not installed: https://www.distributed-ci.io/jobs/761cd8cb-02b9-4e9a-af94-41894a1ed28c/jobStates?sort=date&task=03457f31-5ef4-49da-add2-05141b2cf145
working job, not impacted with change because nmstate is installed: https://www.distributed-ci.io/jobs/b1aacd2f-1f3b-46c6-8df5-80d1b2f2386d/jobStates?sort=date&task=dceb8965-aef8-4fd7-a09e-8cab512e30c6

Red Hat SSO required to see the jobs, he failed job looks like this:
Screenshot from 2023-07-25 18-13-15

PR #257, opened 2023-07-19
Add support to NATted networks

ramperher:
- Currently, only bridged networks are supported for VMs created by Crucible
- There are partners interested in moving to a NATted network instead
- Adding this feature, using bridged networks by default
- Adding instructions about how to use it
- Adding a missed field in nmstate template related to MAC addresses

### DIFF
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -382,7 +382,7 @@ If you need VMs with static UUIDs to allow them to be reused then the UUID for a
 
 #### Bridged network vs. NATted network
 
-By default, this repository deploys a bridged network to interconnect all virtual machines created between them, but it is possible to use a NATted network instead. You just need to set `use_natted_network` to `true` (default `false`) and `use_bridged_network` to `false`(default `true`).
+By default, this repository deploys a bridged network to interconnect all virtual machines created between them, but it is possible to use a NATted network instead. To select one option or another, you need to set up `vm_network_mode` to `bridge` (default) or `nat`, respectively.
 
 ### SSH Key Gen
 
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -380,6 +380,9 @@ If you need VMs with static UUIDs to allow them to be reused then the UUID for a
               uuid: d36ebda0-25bf-55ef-9c69-66ad5ef0d39d
 ```
 
+#### Bridged network vs. NATted network
+
+By default, this repository deploys a bridged network to interconnect all virtual machines created between them, but it is possible to use a NATted network instead. You just need to set `use_natted_network` to `true` (default `false`) and `use_bridged_network` to `false`(default `true`).
 
 ### SSH Key Gen
 
--- a/roles/create_vms/tasks/prepare_network.yml
+++ b/roles/create_vms/tasks/prepare_network.yml
@@ -1,6 +1,16 @@
 - name: Setup network
   become: true
   block:
+    - name: Check if the selected network mode is allowed
+      vars:
+        allowed_vm_network_modes:
+          - bridge
+          - nat
+      assert:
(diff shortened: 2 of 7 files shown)


roles/process_nmstate/templates/nmstate.yml.j2
```diff
@@ -10,6 +10,9 @@ dns-resolver:
 interfaces:
 {% for interface in network_config.interfaces %}
   - name: {{ interface.name }}
+    {% if interface.mac is defined %}
+    mac: {{ interface.mac }}
```
nocturnalastro, line 14: I think its `mac-address` in NMState :)

docs/inventory.md
```diff
@@ -372,6 +372,9 @@ If you need VMs with static UUIDs to allow them to be reused then the UUID for a
               uuid: d36ebda0-25bf-55ef-9c69-66ad5ef0d39d
 ```
 
+#### Bridged network vs. NATted network
+
+By default, this repository deploys a bridged network to interconnect all virtual machines created between them, but it is possible to use a NATted network instead. You just need to set `use_natted_network` to `true` (default `false`) and `use_bridged_network` to `false`(default `true`).
```
nocturnalastro, line 377: What happens if people set both to false by accident, would this completely break the install? I think at very least we need an inventory validation to protect against this. 

And perhaps a single value to switch between the two would be better something like `vm_network_type`. Once again with an inventory validation to check that is set to an allowed value.

nocturnalastro (changes requested):

roles/process_nmstate/templates/nmstate.yml.j2
ramperher: Fixed, I've mainly included it because I've seen several network configs making use of MAC addresses, but the template for nmstate didn't have it.

nocturnalastro: Yeah we use it to populate `mac_interface_map` which the installers use to select the correct nmstate for the host.

roles/create_vms/tasks/prepare_network.yml
```diff
@@ -1,6 +1,16 @@
 - name: Setup network
   become: true
   block:
+    - name: Check if the selected network mode is allowed
```
nocturnalastro, line 4: Could you move this into the `validate_inventory` role that way it will be checked at the start of the execution so people don't have to wait for feedback :)

ramperher: Sure, done

ramperher:
Thanks @nocturnalastro for the review, we'll validate now with the partner to confirm this works, then I'll merge :)

nocturnalastro (approved):

ramperher:
We've stopped the testing with the partner we're addressing this issue to focus on other tasks with more priority, that's why I've not run more tests and checks regarding this. But this is still ongoing, we'll resume the tests to validate this feature in the near future when agreeing that with the partner. So please, do not close/merge this PR yet, we've not forgotten about this :) thanks!

nocturnalastro:
@ramperher If this is ready for merge can you squash your commits :)

nocturnalastro:
Fair enough. I'll add need more testing tag back on then for now. :)

ramperher:
I'll be testing this change here in the collections repo: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/92, since it's easier for us. Then, I'll move the final code here just to have it in the crucible repo, just in case

ramperher:
Closing, moving in favour of https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/92 in ansible-collections-redhatci-ocp repo

Issue #261, opened 2023-08-11
Add support to pass dualstack VIPs for OCP >= 4.12


manurodriguez:
### Feature description

Hi! I'm reading this note in the crucible docs: https://github.com/redhat-partner-solutions/crucible/blob/main/docs/inventory.md#dual-stack
```
Openshift currently only allows the ingress and API VIPs to be single stack so you must choose IPv4 or IPv6. 
```
And I assume that's based on an older Openshift version, since starting in OCP 4.12 we can specify both VIPs using `apiVips` and `ingressVIPs`:
```
$ openshift-install explain installconfig.platform.baremetal
KIND:     InstallConfig
VERSION:  v1

RESOURCE: <object>
  BareMetal is the configuration used when installing on bare metal.

FIELDS:
    apiVIP <string>
      Format: ip
      DeprecatedAPIVIP is the VIP to use for internal API communication Deprecated: Use APIVIPs

    apiVIPs <[]string>
      Format: ip
      APIVIPs contains the VIP(s) to use for internal API communication. In dual stack clusters it contains an IPv4 and IPv6 address, otherwise only one VIP
...
    ingressVIP <string>
      Format: ip
      DeprecatedIngressVIP is the VIP to use for ingress traffic Deprecated: Use IngressVIPs
                      
    ingressVIPs <[]string>
      Format: ip
      IngressVIPs contains the VIP(s) to use for ingress traffic. In dual stack clusters it contains an IPv4 and IPv6 address, otherwise only one VIP
...
```
Currently Crucible has two variables to pass the VIPs values: `api_vip` and `ingress_vip` but each variable is designed to have only one value either an IPv4 or an IPv6 address, not both.

As a crucible user that deploys OCP 4.12 and above, I would like to be able to specify VIPs for both IPv4 and IPv6.

### Required statements

- [X] I acknowledge that Red Hat does not provide commercial support for the content of this repository.
- [X] I acknowledge that any assistance is offered purely on a best-effort basis, as resource permits.

manurodriguez:
Now the challenge is how to do this without affecting much current users that rely on using these variables.
I was thinking two different approaches: 
1. Adding two new variables, i.e.: `api_vip6` and `ingress_vip6` specifically for Ipv6, and make sure to include validations to confirm the old vars are IPv4 and the new vars are IPv6, however that will force folks that currently use `api_vip` and `ingress_vip`  for IPv6 VIPs to change their inventories/deployments.
```
[...]

PR #262, opened 2023-08-14
add extra VIPs support for dualstack on OCP >= 4.12

manurodriguez:
This change allows to pass extra VIPS for dualstack deployments using extra_api_vip and extra_ingress_vip variables.

Fixes: #261

### DIFF
--- a/roles/validate_inventory/tasks/ai.yml
+++ b/roles/validate_inventory/tasks/ai.yml
@@ -28,3 +28,60 @@
     fail_msg: "{{ item }} is not within the machine network!"
   when: vip_dhcp_allocation == false
   loop: "{{ groups['masters'] + (groups['workers'] | default([])) }}"  # This should not include day2_workers as they can be RWNs
+
+- name: Validate extra VIPs for dualstack
+  when:
+    - extra_api_vip is defined
+    - extra_api_vip | length > 0
+    - extra_ingress_vip is defined
+    - extra_ingress_vip | length > 0
+  block:
+    - name: Assert that Openshift version is supported for dualstack VIPs
+      assert:
+        that:
+          - openshift_full_version is version('4.12', '>=')
+        fail_msg: "openshift_full_version  must be >= 4.12. to support dualstack VIPs"
+
+    - name: Assert that extra_machine_networks variable is defined and not empty
+      assert:
+        that:
+          - (extra_machine_networks | length) > 0
+        fail_msg: "extra_machine_networks must be defined and have at least one cidr value in a list to support dualstack VIPs"
+
+    - name: Check if extra api VIP is within the extra machine networks
+      vars:
+        extra_api_vip: "{{ hostvars['assisted_installer']['extra_api_vip'] | default ([]) }}"
+        extra_api_vip_tests: []
+      ansible.builtin.set_fact:
+        extra_api_vip_tests: "{{ extra_api_vip_tests + [extra_api_vip | ansible.utils.ipaddr(item.cidr)] }}"
+      when:
+        - vip_dhcp_allocation == false
+        - extra_machine_networks is defined
+      loop: "{{ hostvars['assisted_installer']['extra_machine_networks'] }}"
+
+    - name: Fail if extra api VIP is NOT within any of the extra machine networks
+      ansible.builtin.assert:
+        that:
+          - extra_api_vip in extra_api_vip_tests
+        fail_msg: "{{ extra_api_vip }} is not within any of the extra machine networks!"
+      when:
+        - extra_api_vip_tests is defined
+
+    - name: Check if extra ingress VIP is within the extra machine networks
+      vars:
+        extra_ingress_vip: "{{ hostvars['assisted_installer']['extra_ingress_vip'] | default ([]) }}"
+        extra_ingress_vip_tests: []
+      ansible.builtin.set_fact:
+        extra_ingress_vip_tests: "{{ extra_ingress_vip_tests + [extra_ingress_vip | ansible.utils.ipaddr(item.cidr)] }}"
+      when:
+        - vip_dhcp_allocation == false
+        - extra_machine_networks is defined
+      loop: "{{ hostvars['assisted_installer']['extra_machine_networks'] }}"
+
+    - name: Fail if extra ingress VIP is NOT within any of the extra machine networks
+      ansible.builtin.assert:
+        that:
+          - extra_ingress_vip in extra_ingress_vip_tests
+        fail_msg: "{{ extra_ingress_vip }} is not within any of the extra machine networks!"
+      when:
+        - extra_ingress_vip_tests is defined
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -271,7 +271,10 @@ To enable assisted installer to communicate via IPv6 you must first have the hos
 
 #### Dual Stack
 
-Openshift currently only allows the ingress and API VIPs to be single stack so you must choose IPv4 or IPv6. Then crucible offers 3 variables for the extra network configuration (`extra_machine_networks`, `extra_service_networks` and `extra_cluster_networks`):
+Crucible offers 3 variables for the extra network configuration (`extra_machine_networks`, `extra_service_networks` and `extra_cluster_networks`), and for the VIPs there are two main configurations:
+
+- Openshift 4.11 and below only allow the ingress and API VIPs to be single stack so you must choose IPv4 or IPv6.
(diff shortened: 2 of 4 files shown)


roles/generate_manifests/tasks/main.yml
```diff
@@ -27,6 +27,20 @@
       }}"
   when: use_local_mirror_registry  | bool
 
+- name: Update api_vips variable when extra_api_vip is defined
+  ansible.builtin.set_fact:
+    api_vips: "{{ api_vips + [extra_api_vip] }}"
+  when:
+    - extra_api_vip is defined
+    - extra_api_vip | length > 0
```
nocturnalastro, line 35: You can do the check for length in the inventory validation. That way simplify your check here. 

This would allow you to set `api_vips` in a relativity  simple jinja template within defaults like we do with dns for the assisted installer deployment [here](https://github.com/redhat-partner-solutions/crucible/blob/main/roles/setup_assisted_installer/defaults/main.yml#L50C1-L50C79). Allowing you to just use `api_vips` as if was natural. Just a suggestion if you would prefer to have it explicitly in tasks that's also fine.

manurodriguez: Thanks, I refined the validations with your suggestions and looks simpler this way. I launched a test to confirm this is non breaking current clusters without dualstack and looks good, but I still have pending to test again in a baremetal cluster with dualstack. I'll keep you posted.

roles/validate_inventory/tasks/ai.yml
```diff
@@ -28,3 +28,36 @@
...
+    - name: Assert that Openshift version is supported for dualstack VIPs
+      assert:
+        that:
+          - openshift_full_version is version('4.12', '>=')
+        fail_msg: "openshift_full_version  must be >= 4.12. to support dualstack VIPs"
+
+    - name: Assert extra api VIP is within the extra machine networks
+      assert:
+        that:
+          - hostvars['assisted_installer']['extra_api_vip'] | ansible.utils.ipaddr(item.cidr) | ansible.utils.ipaddr('bool')
```
nocturnalastro, line 43: I don't think this is quite right. This is checking if the `api_vip` is within all the networks where as it should be in any of the extra networks. So I think you need to collect the tests into a list and then check the list with `is any` so that you don't false negatives.

roles/validate_inventory/tasks/ai.yml
```diff
@@ -28,3 +28,36 @@
...
+    - name: Assert extra ingress VIP is within the extra machine networks
+      assert:
+        that:
+          - hostvars['assisted_installer']['extra_ingress_vip'] | ansible.utils.ipaddr(item.cidr) | ansible.utils.ipaddr('bool')
+        fail_msg: "{{ extra_ingress_vip }} is not within the extra machine networks!"
+      when:
+        - vip_dhcp_allocation == false
+        - extra_machine_networks is defined
+      loop: "{{ hostvars['assisted_installer']['extra_machine_networks'] }}"
+  when:
```
nocturnalastro, line 59: Can you put `when` above `block` it makes the block easier to read.

nocturnalastro (changes requested):

roles/validate_inventory/tasks/ai.yml
manurodriguez: Done, added an extra commit

manurodriguez:
Last changes also tested on Baremetal with dualstack in: https://www.distributed-ci.io/jobs/1c452a53-61cc-4fdf-b51a-0d0f1337c3c9/jobStates

roles/validate_inventory/tasks/ai.yml
manurodriguez: Thanks for the clarification, I think I have a better idea now, we should only fail the validation if none an extra vip is not in any of the `extra_machine_networks`,  if it matches one, it should pass right? 

I updated the PR, and I included some tests of the conditions [here](https://paste.openstack.org/show/bnblDSR3dt72xO80UvTT/)

roles/validate_inventory/tasks/ai.yml
```diff
@@ -28,3 +28,56 @@
...
+    - extra_ingress_vip | length > 0
+  block:
+    - name: Assert that Openshift version is supported for dualstack VIPs
+      assert:
+        that:
+          - openshift_full_version is version('4.12', '>=')
+        fail_msg: "openshift_full_version  must be >= 4.12. to support dualstack VIPs"
+
+    - name: Check if extra api VIP is within the extra machine networks
+      ansible.builtin.debug:
```
nocturnalastro, line 46: just use `set_fact` here, if people need to see it they can use the vebosity flag to see the results of the task. 

```yaml
- name:  Check if extra api VIP is within the extra machine networks
  vars:
     extra_api_vip: "{{ hostvars['assisted_installer']['extra_api_vip'] }}"
  ansible.builtin.set_fact:
     extra_api_vip_tests: "{{ extra_api_vip_tests + [extra_api_vip | ansible.utils.ipaddr(item.cidr)]"
  loop: "{{ hostvars['assisted_installer']['extra_machine_networks'] }}"
```

roles/validate_inventory/tasks/ai.yml
```diff
@@ -28,3 +28,56 @@
...
+    - extra_api_vip | length > 0
+    - extra_ingress_vip is defined
+    - extra_ingress_vip | length > 0
+  block:
+    - name: Assert that Openshift version is supported for dualstack VIPs
+      assert:
+        that:
+          - openshift_full_version is version('4.12', '>=')
+        fail_msg: "openshift_full_version  must be >= 4.12. to support dualstack VIPs"
+
```
nocturnalastro, line 44: Can you also add a test that `(extra_machine_networks | length) > 0`
It will be a good validation as having it empty wouldn't be valid anyway and also stops a failure case of the following tests.

nocturnalastro (changes requested):
Just some small tweaks then I think we're good here :) Thanks for your work :)

roles/validate_inventory/tasks/ai.yml
```diff
@@ -28,3 +28,56 @@
...
+      register: extra_api_vip_tests
+      loop: "{{ hostvars['assisted_installer']['extra_machine_networks'] }}"
+
+    - name: Fail if extra api VIP is NOT within any of the extra machine networks
+      vars:
+        vip_field_filter: "results[*].msg"
+        extra_api_ipv6_list: "{{ extra_api_vip_tests | json_query(vip_field_filter) }}"
+      ansible.builtin.assert:
+        that:
+          - extra_api_vip in extra_api_ipv6_list
```
nocturnalastro, line 60: Why does this have ipv6 in the name, is there an ordering requirement on the vips? Does it have to go ipv4 then ipv6? If so can you add a check for this ordering otherwise people might put them in the wrong order and not find out until lots of things have happened.

roles/validate_inventory/tasks/ai.yml
manurodriguez: done, I squashed all commits and push latest changes in one.

roles/validate_inventory/tasks/ai.yml
manurodriguez: sure, done

roles/validate_inventory/tasks/ai.yml
manurodriguez: not really, I was having in mind ipv6 for my tests all this time, so I renamed the variable to be independent of any ipv4 or ipv6

nocturnalastro (approved):

manurodriguez:
Last CI job failed because I was testing a deployment without dualstack (no `extra_*` variables), but hopefully this helped to spot an issue in this validation, despite the block has a condition to check if `extra_api_vip` is defined and non-empty, and none of the tasks inside should be performed, looks like the vars in the tasks are interpreted.
```
      vars:
-        extra_api_vip: "{{ hostvars['assisted_installer']['extra_api_vip'] }}"
+        extra_api_vip: "{{ hostvars['assisted_installer']['extra_api_vip'] | default ([]) }}"

```
I amended the commit adding a default value. And launching new tests again

manurodriguez:
The last test validated we are not breaking deployments without dualstack, now it is pending to test a deployment with dualstack and extra VIPs.

manurodriguez:
FYI: We finally tested in a dualstack baremetal cluster with latest OCP 4.12, and we confirmed all the VIPs are there:
https://www.distributed-ci.io/jobs/792daffb-5e1b-4118-b6be-99590434fce8/jobStates?sort=date

arjuhe (approved):
+1 LGTM

PR #271, opened 2023-09-14
Revert changes to use community.crypto collection in setup_selfsigned_cert

ramperher:
This reverts some of the changes included in https://github.com/redhat-partner-solutions/crucible/commit/c86104c6b1b6070043c4eccb9f18993430412484

While testing Assisted deployments with DCI on a daily basis, e.g. here: https://www.distributed-ci.io/jobs/68c4cb1f-0ff6-4a76-b6f0-4c2704c78214/jobStates?sort=date, we discovered that the usage of community.crypto is affecting the jobs, since it's not available for downloading through RPM. Its usage in DCI would depend on creating a custom RPM and inject it in the project, and that's something we would like to avoid unless it's unavoidable to use community.crypto.

Then, the proposal is to revert the usage of community.crypto in setup_selfsigned_cert.

### DIFF
--- a/roles/setup_selfsigned_cert/tasks/main.yml
+++ b/roles/setup_selfsigned_cert/tasks/main.yml
@@ -31,12 +31,11 @@
   become: true
 
 - name: Generate an OpenSSL private key
-  community.crypto.openssl_privatekey:
+  openssl_privatekey:
     path: "{{ privatekey_path }}"
 
-
 - name: Generate an OpenSSL CSR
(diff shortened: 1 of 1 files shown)


nocturnalastro (approved):

ramperher:
Change worked! @nocturnalastro I think this can be merged, but I can't do it till having two approving reviews

nocturnalastro:
@ramperher I'll merge it aldo was fine with the revert and you've reverted :)

PR #282, opened 2023-10-27
Update ai_verion to latest

nocturnalastro:


gojeaqui:
@nocturnalastro beware with this change because it is not as easy as changing the version, look at the issue #288.

Updating `ai_version` without testing might leave the repo unusable

nocturnalastro:
@gojeaqui The API doesn't tend to change drastically in practice but be we always run though a set of deployments before bumping the version that's why this hasn't been merged yet :)

PR #283, opened 2023-10-27
vm_network_test_ip parameter addition at inventory.yml

ldornele:
vm_network_test_ip parameter addition at inventory.yml due to task : Check if vm_host is bastion


nocturnalastro (approved):
Thanks :)

PR #287, opened 2023-11-10
Change some checks to acommodate for clients where ICMP is disabled by firewall

gojeaqui:
When testing for connectivity Use NetCat instead of ping to test BMC and NTP connection, since some firewalls might block ICMP.
On the ohter hand the "bmc_ip" variable should not be required for `vendor: PXE`


nocturnalastro:
Theres another ping for testing network connectiveity after appling the network config. Could you update this one as well. 
https://github.com/redhat-partner-solutions/crucible/blob/main/roles/apply_nmstate/tasks/main.yml#L40

gojeaqui:
> Theres another ping for testing network connectiveity after appling the network config. Could you update this one as well. https://github.com/redhat-partner-solutions/crucible/blob/main/roles/apply_nmstate/tasks/main.yml#L40

@nocturnalastro in this case it is OK to have a ping because that IP address does not correspond to any type of service so I wouldn't know to which port NetCat should probe. Besides it is configurable to any host the user chooses.

PR #290, opened 2023-11-21
introduce variable  hosts_discovery_minutes

newkit:
in order to make host discovery timeout configurable

### DIFF
--- a/roles/install_cluster/defaults/main.yml
+++ b/roles/install_cluster/defaults/main.yml
@@ -27,3 +27,6 @@ host_discovery_states:
 cluster_installation_states:
   - installing
   - installing-pending-user-action
+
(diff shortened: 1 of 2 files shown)


nocturnalastro (changes requested):
Can you move this into the defaults file that the default only in one place

newkit:
@nocturnalastro Do you mean like this?

nocturnalastro (approved):
Yeah, like that there is only one place to change the defaults which makes life a lot easier and the tasks just look cleaner :)

nocturnalastro:
@newkit Could squash your commits for me then I'll get it merged :)

newkit:
@nocturnalastro All commits are squashed into one now.

PR #296, opened 2023-12-07
Abi testing rhel9

arjuhe:
- Modified: roles/generate_agent_iso/tasks/main.yml
  - Added a become to "Create podman auth dir".  
- roles/generate_manifests/templates/registry-config.j2
  - Removed some templating that was changing `imageContentSources` into `imageDigesstSources`.


arjuhe:
Accidental PR request. Closed

PR #299, opened 2023-12-15
Update XDG_RUNTIME_DIR path in generate_agent_iso

nocturnalastro:
This was causing DCI's abi tests to fail. See here:
https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/93


nocturnalastro:
No longer needed as we're now using the collection

